PR #3, opened 2026-01-08
Initial donation from embrace.io

fractalwrench:
## Goal

Initial donation of the existing codebase from https://github.com/embrace-io/opentelemetry-kotlin/commit/0217fb7c51134c5156f52f1d7c4c11d4390a7162. I've simply copied all files over that were in the repository at this time. I figured that would be the easiest way of starting, and that we can follow up changes to CI workflows/READMEs etc in subsequent changesets.


Msksgm (approved):
LTGM

PR #17, opened 2026-01-12
Remove opentelemetry-kotlin prefix from modules

### LINKED ISSUE #9: opentelemetry-kotlin module prefix is mega verbose
All (most?) of the modules in this repo have the `opentelemetry-kotlin` prefix in them, which is unnecessarily verbose for a project that is now hosted within the open-telemetry org and in repo named `opentelemetry-kotlin`. 

Changing this will touch nearly all of the code in the project, so we could get it over with earlier by doing it all in one go, or we could do it module-by-module. I'm open to suggestions/preferences.

fractalwrench:
## Goal

Removes the `opentelemetry-kotlin` prefix from modules. This closes #9.


semconv/README.md
```diff
@@ -1,4 +1,4 @@
-# opentelemetry-kotlin-semconv
+# semconv
```
breedx-splk: oh yeah huh I guess I had missed that this was in here. Maybe one day it peels out to its own repo like the java ones? That's a conversation for another time.

breedx-splk (approved):
That's a lot of files!

I cruised through (in the interest of getting this done sooner rather than later) and it seems fine. I wonder if we'll end up bike shedding on a few names (core -> common, for example??) but that's a concern for a later time.

Thanks for doing this early.

PR #50, opened 2026-01-15
Implement `onEnding()` in `SpanProcessor`

### LINKED ISSUE #33: Implement OnEnding function in SpanProcessor
Implement `onEnding` that is called in `SpanProcessor` when a `Span` is about to end: https://opentelemetry.io/docs/specs/otel/trace/sdk/#onending

fractalwrench:
## Goal

Implements [`onEnding()` in `SpanProcessor`](https://opentelemetry.io/docs/specs/otel/trace/sdk/#onending). This allows for a callback to be invoked just before a `Span` is about to end.

Closes #33

## Testing

Added unit tests.

### DIFF
--- a/exporters-otlp/build.gradle.kts
+++ b/exporters-otlp/build.gradle.kts
@@ -5,6 +5,15 @@ plugins {
     id("signing")
     id("com.vanniktech.maven.publish")
     id("org.jetbrains.kotlinx.kover")
+    alias(libs.plugins.buildKonfig)
+}
+
+buildkonfig {
+    packageName = "io.opentelemetry.kotlin.export"
+
+    defaultConfigs {
+        buildConfigField(com.codingfeline.buildkonfig.compiler.FieldSpec.Type.STRING, "VERSION", project.version.toString())
+    }
 }
 
 kotlin {
@@ -35,4 +44,8 @@ kotlin {
             }
         }
     }
+}
+
+tasks.withType<io.gitlab.arturbosch.detekt.Detekt>().configureEach {
+    exclude { it.file.path.contains("buildkonfig") }
 }
--- a/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpClient.kt
+++ b/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpClient.kt
@@ -3,10 +3,12 @@ package io.opentelemetry.kotlin.export
 import io.ktor.client.HttpClient
 import io.ktor.client.plugins.HttpRequestTimeoutException
 import io.ktor.client.plugins.compression.compress
+import io.ktor.client.request.header
 import io.ktor.client.request.post
 import io.ktor.client.request.setBody
 import io.ktor.client.statement.bodyAsBytes
 import io.ktor.http.ContentType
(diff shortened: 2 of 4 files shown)


Msksgm:
@fractalwrench

Hello, it looks like the Codecov CI check for this PR is failing.  
From what I can see, in order to fix this we would need an additional test that covers the nullable processor branch in `implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt`.

https://github.com/open-telemetry/opentelemetry-kotlin/pull/50/changes#diff-b19cfd0aefc8e6996fb99f4671f79b4bd0e82292fae7c9a25610aa2346d504c8R88

I noticed that `IntegrationTestHarness` always adds a non-null processor when it is created.  
So to cover this branch and improve test coverage, we would either need to add a dedicated `SpanModelTest` like the one below, or update the `IntegrationTestHarness` properties so that it can initialize a nullable processor.

https://github.com/Msksgm/opentelemetry-kotlin/commit/97f06e05ca6312426da47bebe789cf4531845e70

```kotlin
package io.opentelemetry.kotlin.integration.test.tracing

import io.opentelemetry.kotlin.ExperimentalApi
import io.opentelemetry.kotlin.InstrumentationScopeInfoImpl
import io.opentelemetry.kotlin.clock.FakeClock
import io.opentelemetry.kotlin.resource.FakeResource
import io.opentelemetry.kotlin.tracing.FakeSpanContext
import io.opentelemetry.kotlin.tracing.fakeSpanLimitsConfig
import io.opentelemetry.kotlin.tracing.model.SpanKind
import io.opentelemetry.kotlin.tracing.model.SpanModel
import kotlin.test.Test
import kotlin.test.assertTrue

@OptIn(ExperimentalUnsignedTypes::class)
internal class SpanModelTest {

    @OptIn(ExperimentalApi::class)
    @Test
    fun testEndWithNullProcessor() {
        val spanModel = SpanModel(
            clock = FakeClock(),
            processor = null,
            name = "test-span",
            spanKind = SpanKind.INTERNAL,
            startTimestamp = 0L,
            instrumentationScopeInfo = InstrumentationScopeInfoImpl("test", null, null, emptyMap()),
            resource = FakeResource(),
            parent = FakeSpanContext.INVALID,
            spanContext = FakeSpanContext.VALID,
            spanLimitConfig = fakeSpanLimitsConfig
        )

        spanModel.end()

        assertTrue(spanModel.hasEnded)
    }
}

```

I have two questions:

1. Do we want to satisfy the Codecov coverage check in this repository, or is it acceptable to leave this particular branch uncovered? 
2. Are there any existing testing guidelines (for example, around the use of test harnesses) that we should follow for cases like this?

fractalwrench:
@Msksgm IMO yes, we should always try to satisfy the code coverage checks in the repository. I think it'd be acceptable to ignore them in a case where it's not possible to test code (e.g. if a new interface without any implementation is being added). I've added some extra coverage to this changeset that tests `SpanModel` in isolation.

There aren't existing guidelines for this case. So far I've tended to prefer tests that use `IntegrationTestHarness` as they tend to focus on the desired functionality compared to a more granular unit test that tends to be coupled to the implementation details. That very much depends on the scenario though - this case seems easier to verify with a unit test.

PR #55, opened 2026-01-16
Add syntactic sugar for wrapping operations with a span

### LINKED ISSUE #4: Make SpanModel implement AutoCloseable
Originally raised in the old repository: https://github.com/embrace-io/opentelemetry-kotlin/issues/357

>The Kotlin stdlib defines AutoCloseable from 2.0 onwards. This acts similarly to the JVM's AutoCloseable and allows resources to be cleaned up. It would be desirable to have `Span` and possibly `Context` (when attaching then detaching in an execution unit) implement this interface, as there are IDE hints associated with AutoCloseable that prompt end-users to clean up the resource.

>I think this will require bumping our min supported Kotlin version to 2.0 however.

fractalwrench:
## Goal

Adds syntactic sugar to make it easier for library consumers to wrap operations with a `Span`. This works by providing a [higher-order function](https://kotlinlang.org/docs/lambdas.html#higher-order-functions) that ends the span after its execution is ended.

This has the added bonus of:

1. Requiring a status to be set when ending the span
2. Automatically [recording exceptions](https://opentelemetry.io/docs/specs/otel/trace/exceptions/#recording-an-exception) as events when they happen in the operation
3. Automatically setting the status to `Error` if the operation throws an exception

Originally the approach suggested `AutoCloseable` but I feel this approach is preferable given that it's now in an opt-in module of syntactic enhancements.

Closes #4 

## Testing

Added unit tests.


bidetofevil (approved):
Nice. LGTM.

Msksgm (approved):
I asked a question.
However, I also think that it is not a big issue.
Then, approved.

PR #58, opened 2026-01-16
Implement stdout exporters

### LINKED ISSUE #38: Create stdout exporters
Create built-in exporters for logs and traces that log to stdout: https://opentelemetry.io/docs/specs/otel/trace/sdk_exporters/stdout/

fractalwrench:
## Goal

Implements log/span exporters that print to stdout. This defaults to `println` but the actual destination is configurable via the API to create the exporter.

Closes #38.

## Testing

Added unit tests.


Msksgm (approved):
LGTM.

However, should we add a warning above the stdout log exporter or stdout trace exporter?

The specification for the stdout trace exporter recommends including wording like the following:
https://opentelemetry.io/docs/specs/otel/trace/sdk_exporters/stdout/

> This exporter is intended for debugging and learning purposes. It is not recommended for production use. The output format is not standardized and can change at any time.
>
> If a standardized format for exporting traces to stdout is desired, consider using the [File Exporter](https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/), if available. However, please review the status of the File Exporter and verify if it is stable and production-ready.

PR #68, opened 2026-01-22
Alter `export()` function so that it uses suspend modifier

### LINKED ISSUE #5: Parity between OperationResultCode and CompletableResultCode for deferred results
Originally raised in the old repository: https://github.com/embrace-io/opentelemetry-kotlin/issues/445

>There seems to be a notable gap in the expressiveness of [`OperationResultCode`](https://github.com/embrace-io/opentelemetry-kotlin/blob/9b11da7045a60504bd33160777ad8b19f9fa9e68/opentelemetry-kotlin-api/src/commonMain/kotlin/io/embrace/opentelemetry/kotlin/export/OperationResultCode.kt#L9) compared to OTel Java's [`CompletableResultCode`](https://github.com/open-telemetry/opentelemetry-java/blob/da295cfa3c93b73376e4c32f2dd450b2035483b3/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java#L26), particularly around accommodation of deferred results. Curious to hear from the maintainers on whether there is a plan to address this parity.

>One place where this becomes apparent is in [`TelemetryExporter#export(List<T>)`](https://github.com/embrace-io/opentelemetry-kotlin/blob/9b11da7045a60504bd33160777ad8b19f9fa9e68/opentelemetry-kotlin-exporters-otlp/src/commonMain/kotlin/io/embrace/opentelemetry/kotlin/export/TelemetryExporter.kt#L25) as `Success` is immediately returned regardless of whether there is a downstream failure. Retrial mechanism notwithstanding, this could provide a false impression on the state of export to the client.

fractalwrench:
## Goal

Alters the `export()` function of the `SpanExporter` and `LogRecordExporter` so that they use the `suspend` modifier. This gives syntax similar to `CompleteableResultCode` and therefore closes #5.

## Testing

Updated existing unit tests to use [`runTest`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/run-test.html).


bidetofevil (approved):
LGTM

Issue #69, opened 2026-01-22
Prevent data loss during process termination when telemetry is not exported


fractalwrench:
## Goal

Currently data is lost if it not exported at the point of process termination. This is an important use-case for mobile because processes can be terminated at any time and network connectivity can be intermittent. It is therefore necessary to persist telemetry and continue its export on the next process launch. This project should provide APIs that facilitate this behavior.

### Background

opentelemetry-kotlin currently does not handle data loss in several scenarios:

- JVM crash
- C signal
- OS signals
- No network connectivity

Each scenario has its own constraints. For a JVM crash it’s possible to set a handler. For an NDK crash a handler is available but all code must be async-safe. When the OS terminates the process there is no mechanism to respond. When there’s no network connectivity, the process will eventually terminate and cause data loss.

### Scope of problem

The Logging and Tracing APIs are the main signals we need to persist. Logs are one point-in-time so only need persisting once. Traces are more complicated as a span can be updated after creation, and may not end. For now, we'll only consider how to persist a span after it's ended. Metrics is also out-of-scope as it’s not implemented yet in opentelemetry-kotlin.

This design is written primarily with the JVM in mind although the approach should be broadly the same on any platform with filesystem access.

## Suggested approach

First, some background on how export works. Currently `TracerProvider` has a single `CompositeSpanProcessor`, that delegates calls to other `SpanProcessor` objects.

A `SpanProcessor` *may* have a `SpanExporter` that exports values, but that is not mandatory. For example, a `SpanProcessor` can simply append an attribute such a session ID to a span. The next processor in the chain would then have responsibility for exporting.

Therefore, we want data persistence to happen at the *end* of the processing chain but the *beginning* of the export chain. This is possible if we wrap all other processors and exporters declared for an application.

### Telemetry persistence approach

The suggested approach only considers persisting telemetry at a single point-in-time. `LogRecord` is therefore the example, although this equally applies to `Span` after `onEnd()` is invoked on the `SpanProcessor`. The high-level pipeline would look like this:

1. Final `LogRecord` is emitted
2. `SimpleLogRecordProcessor#onEmit` invoked
3. `PersistingLogRecordExporter#onExport` invoked
4. `List<ReadableLogRecord>` iterated over and a file is created for each telemetry object. The telemetry is stored as an encoded binary protobuf
5. The filename encodes the observed timestamp, a UUID, and the telemetry type (e.g. ‘log’)
[...]

PR #88, opened 2026-01-27
Create `PersistedTelemetryRecord` class

fractalwrench:
## Goal

Creates `PersistedTelemetryRecord` and `PersistedTelemetryConfig`. This changeset aims to start addressing #69 by creating a new opt-in exporter that writes telemetry to disk.

These classes will be used to encode information in the filename about the stored telemetry, and to limit the amount of telemetry that is stored on disk. Future changesets will add exporters that actually use these, amongst other features.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. Looking forward to seeing how this is leveraged.

PR #92, opened 2026-01-28
Support writing telemetry to file system

fractalwrench:
## Goal

Supports writing telemetry to the file system using Okio. Telemetry is stored gzip compressed on JVM/Android/iOS platforms, and is stored as-is on JS.

This will be used for persisting telemetry on disk in an exporter to prevent data loss in the case of process termination.

## Testing

Added unit tests with Okio's [fake file system](https://square.github.io/okio/3.x/okio-fakefilesystem/okio-fakefilesystem/okio.fakefilesystem/-fake-file-system/).

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -28,7 +28,7 @@ jobs:
           java-version: 21
 
       - name: Setup Gradle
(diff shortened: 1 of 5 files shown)


bidetofevil (approved):
Very nice start. I assume there will be more resilience built on a higher layer that is consuming this to deal with unexpected terminations?

fractalwrench:
@bidetofevil correct, that's the plan

PR #102, opened 2026-02-02
Delete telescope app

fractalwrench:
## Goal

Deletes 'telescope-app'. This is an Android app in a standalone project within the repository and as such:

- Doesn't provide a README with instructions on how to run it like the other example apps do
- Doesn't work out-of-the-box because it requires snapshot publishing
- Does not appear to be configured for dependency updates
- Adds a [considerable amount to the build time](https://github.com/open-telemetry/opentelemetry-kotlin/actions/runs/21592419013/job/62215691582) (1m28s compared to 11s for all the other apps)

It's also Android-specific and our example apps should ideally be platform-agnostic and work on any platform.

I therefore propose deleting it in favour of adding future sample code to `:example-common`, where it can be used in a demonstration for all the example apps.


fractalwrench:
I'm definitely leaning towards deleting it and reworking the existing apps as a KMP app (potentially with shared UI) as that would reduce the amount of times we'd have to write example code for individual platforms.

PR #105, opened 2026-02-03
Add AtomicLong implementation

fractalwrench:
## Goal

Adds a multiplatform implementation of `AtomicLong` as Kotlin doesn't have a multiplatform implementation that goes all the way back to 2.0 (although it does have an experimental implementation [starting in 2.1](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.concurrent.atomics/-atomic-long/)). I also considered [atomic-fu](https://github.com/Kotlin/kotlinx-atomicfu) but it's not stable and could change API at any time.

This was primarily driven by wanting a counter that could be incremented atomically without a lock. I've worked around this in a couple of PRs but it would be good to have the option in our toolset.

## Testing

Added unit tests.


bidetofevil (approved):
Very nice. Do we need an implementation for iOS as well?

Issue #73, opened 2026-01-22
When no implementation is provided for the API it should default to no-op


fractalwrench:
## Goal

When no implementation is present the API should default to the no-op implementation. This is outlined in the specification here: https://opentelemetry.io/docs/specs/otel/library-guidelines/#expected-usage

PR #130, opened 2026-02-06
Alter no-op API and enhance docs

fractalwrench:
## Goal

Alters the no-op API from `createNoopOpenTelemetry` to `NoopOpenTelemetry`. IMO this is a better API because we're not creating a new instance each time a function is called, and there is precedence from other Kotlin APIs in the ecosystem for a singleton similar to this.

I've also added some docs to explain what behavior should look like when a valid implementation is not installed. As [suggested in the spec](https://opentelemetry.io/docs/specs/otel/library-guidelines/#expected-usage), the API should default to a no-op instance. IMO the best way of achieving this is for a library author or opentelemetry-kotlin user to inject the `OpenTelemetry` value and default to the noop instance. This certainly feels preferable to a global singleton with mutable state, which is the alternative.


fractalwrench:
Closes #73

PR #148, opened 2026-02-13
Allows using compatibility mode in example app

fractalwrench:
## Goal

Alters the example app to allow using the compat mode in the example app for JVM/Android. This is stacked on top of #144 - the relevant changes are in the last commit.


bidetofevil (approved):
LGTM

PR #153, opened 2026-02-17
Alter API for creating exporters

fractalwrench:
## Goal

Alters the API used to create processors and exporters. The current API looks like this:

```kotlin
val config: OpenTelemetryConfigDsl.() -> Unit = {
    tracerProvider {
        addSpanProcessor(createSimpleSpanProcessor(createStdoutSpanProcessor()))
    }
}
```

Whereas the new API looks like this:

```kotlin
val config: OpenTelemetryConfigDsl.() -> Unit = {
    tracerProvider {
        export {
            simpleSpanProcessor(stdoutSpanProcessor())
        }
    }
}
```

There are several motivations behind this attempt:

1. This improves API discoverability as the functions are annotated with `@ConfigDsl` so will show up in IDE auto-completions
2. The API more closely matches how the rest of OpenTelemetry is configured & is less verbose
3. It makes the processor/exporter ordering somewhat more explicit

I have not converted all the existing APIs to use this approach yet as I wanted feedback first.

This doesn't break API compatibility yet as I've deprecated the old symbols and provided migration steps via `ReplaceWith`, although we should remove the old symbols sooner rather than later IMO.

## Testing

Added unit tests.


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/OpenTelemetryImplEntrypoint.kt
```diff
@@ -11,8 +11,20 @@ import io.opentelemetry.kotlin.tracing.TracerProviderImpl
  * Constructs an [OpenTelemetry] instance that uses the opentelemetry-kotlin implementation.
  */
 @ExperimentalApi
-public fun createOpenTelemetry(config: OpenTelemetryConfigDsl.() -> Unit = {}): OpenTelemetry {
+public fun createOpenTelemetry(
+
+    /**
+     * Defines the [Clock] implementation used by OpenTelemetry.
+     */
+    clock: Clock = ClockImpl(),
```
fractalwrench: Moving `clock` out of `OpenTelemetryConfigDsl` was another driver behind this change, as it was unwieldy for folks to get the `clock` implementation in the processor/exporter if required.

PR #155, opened 2026-02-18
Alter location of hex on TraceFlags

fractalwrench:
## Goal

Moves `hex` on `TraceFlags` to `api-ext` as it's not part of the API package in the OpenTelemetry specification.

## Testing

Updated existing unit tests.


breedx-splk (approved):
Seems like a nice footprint reduction.

PR #156, opened 2026-02-18
Rename createSpan to startSpan

fractalwrench:
## Goal

Renames `createSpan` to `startSpan` as this better pairs with the `end()` concept, and is also the convention that opentelemetry-java follows.

## Testing


implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/integration/test/tracing/SpanProcessNaughtyOnEndTest.kt
```diff
@@ -32,7 +32,7 @@ internal class SpanProcessNaughtyOnEndTest {
     @Test
     fun testSpanProcessorOnEndOverrideIgnored() = runTest {
         harness.config.spanProcessors.add(NaughtySpanProcessor())
-        harness.tracer.createSpan("span") {
+        harness.tracer.startSpan("span", null, SpanKind.INTERNAL, null) {
```
breedx-splk: parameter names to improve readability

breedx-splk (approved):
This is a nice change. Definitely like the consistency with other impls here!

PR #158, opened 2026-02-19
Rename sdk module as sdk-api

fractalwrench:
## Goal

Renames the `sdk` module as `sdk-api` following discussion in the SIG.


bidetofevil (approved):
LGTM

PR #160, opened 2026-02-19
Alter attributes API to comply with specification

fractalwrench:
## Goal

Alters the attributes API to better comply with the specification. Effectively, this renames `AttributeContainer` to `ReadableAttributes` and `MutableAttributeContainer` to `WritableAttributes`. 

I've also created an internal class named `AttributesModel` which acts as the source of truth. Effectively, these interfaces provide views onto the model, which allows us to control when attributes are read/written. This follows a similar pattern to `SpanModel/ReadableSpan/ReadWriteSpan`, and is something I'd like to attempt more broadly over the API.

I haven't moved any files out of the `api` module to keep the diff small, but will move `WritableAttributes` to `sdk-api` in a future changeset.

## Testing


bidetofevil (commented):
Right. I missed the fact that `WritableAttributes` no longer extend `ReadableAttributes`.

Well, I don't mind the split at the lower-level so long as we can still make the main span interface extend readable so we can arbitrarily look at its state beyond just at the SpanProcessor events level. 

One reason for this is because for the end-user facing app use case, spans can run a long time, and being able to take a snapshot of it an arbitrary time allows us to recover from unexpected crashes. 

So how strict is this edict? I know for backend, the app reading spans is a no-no given the desire to decouple instrumentation code from app code, but is there leeway to open this up a bit given the differing primary use case?

If the default Span being readable is too much, is there another way we can accomplish what I want with additional friction so folks don't accidentally stumble into this pattern?

fractalwrench:
Further context from @carlosalberto would be helpful here. I do think that introspection of attributes and other values on long-running spans is an important consideration. This PR simply follows the letter of the law in its approach.

>is there another way we can accomplish what I want with additional friction so folks don't accidentally stumble into this pattern?

A couple of alternatives I can think of:

1. Cast the reference to `WritableAttributes` (not great)
2. Move this behavior to `api-ext` somehow, so that it's opt-in

sdk-api/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/ReadWriteSpan.kt
```diff
@@ -1,9 +1,10 @@
...
 
 import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.attributes.ReadableAttributes
 
 /**
  * A writable representation of a [Span] that can be modified.
  */
 @ExperimentalApi
-public interface ReadWriteSpan : Span, ReadableSpan
+public interface ReadWriteSpan : Span, ReadableSpan, ReadableAttributes
```
breedx-splk: I think I understand that `ReadableAttributes` is just here to provide the getter on the span, but it feels a little weird that the `ReadWriteSpan` "isA" `ReadableAttributes` instead of `hasA`. Might be a little pedantic, sure, but I think it's important to consider.

It also feels like maybe this should move to the `ReadableSpan` -- so that you get get the read-only version of the attributes from the readable span?

Issue #129, opened 2026-02-06
Make processors/exporters properly respect graceful shutdown.


fractalwrench:
## Goal

Assess existing processors/exporters to check if they respect graceful shutdown, and update implementations if not.

Partially addresses #71.

Issue #128, opened 2026-02-06
Assert telemetry not accepted after shutdown


fractalwrench:
## Goal

Write test cases that check telemetry submitted after shutdown is not accepted. `CloseableOpenTelemetryImplTest` has made a start at this.

Partially addresses #69.

bidetofevil:
Work here will include:
- Make the span and log exporters and processors not process or export telemetry post shutdown
- Make tracer and logger provider provide no-op implementations post shutdown
- Make tracers create no-op spans post shutdown
- Make loggers not emit any telemetry post shutdown
- Tests for all test scenarios

PR #162, opened 2026-02-20
add ShutdownState abstract class and MutableShutdownState

bidetofevil:
## Summary
- Phase 1 of 10: Shutdown compliance stack
- Adds `ShutdownState` abstract class with inline helper methods (`ifActiveOrElse`, `ifActive`, `execute`) and `MutableShutdownState` concrete class with `@Volatile` boolean flag
- Read-only base class supports suspend lambdas via inline members on abstract class

## Stack
1. add ShutdownState abstract class and MutableShutdownState → `main`
2. add shutdown compliance to BatchTelemetryProcessor and TelemetryExporter
3. add shutdown compliance to span processors and exporters
4. add shutdown compliance to log processors and exporters
5. add shutdown compliance to in-memory exporters
6. add shutdown compliance to persistence exporters
7. add shutdown compliance to Java interop adapters
8. propagate read-only ShutdownState to TracerProviderImpl and TracerImpl
9. propagate read-only ShutdownState to LoggerProviderImpl and LoggerImpl
10. wire shared MutableShutdownState through CloseableOpenTelemetryImpl

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sdk-api/api/jvm/sdk-api.api
+++ b/sdk-api/api/jvm/sdk-api.api
@@ -6,6 +6,12 @@ public abstract interface class io/opentelemetry/kotlin/OpenTelemetrySdk : io/op
 	public abstract fun getClock ()Lio/opentelemetry/kotlin/Clock;
 }
 
+public final class io/opentelemetry/kotlin/export/MutableShutdownState : io/opentelemetry/kotlin/export/ShutdownState {
+	public fun <init> ()V
+	public fun isShutdown ()Z
+	public final fun shutdown ()V
+}
+
 public abstract class io/opentelemetry/kotlin/export/OperationResultCode {
 }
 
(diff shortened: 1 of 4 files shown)


bidetofevil:
This stack of PRs should take care of #128 and #129 and then some. I'll see if other tickets are handled partially or fully. I'll also be looking to use `MutableShutdownState` in how `CloseableOpenTelemetryImpl` is implemented once this is merged.

sdk-api/api/jvm/sdk-api.api
```diff
@@ -6,6 +6,12 @@ public abstract interface class io/opentelemetry/kotlin/OpenTelemetrySdk : io/op
 	public abstract fun getClock ()Lio/opentelemetry/kotlin/Clock;
 }
 
+public final class io/opentelemetry/kotlin/export/MutableShutdownState : io/opentelemetry/kotlin/export/ShutdownState {
```
fractalwrench: I don't like that this is exported as part of the `sdk-api` package as it is an implementation detail rather than an API. Is there a more appropriate module such as `exporters-core` or similar that could hold it?

sdk-api/src/commonMain/kotlin/io/opentelemetry/kotlin/export/MutableShutdownState.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import kotlin.concurrent.Volatile
+
+/**
+ * Non-locking but thread-safe implementation of [ShutdownState]. Objects that can read but not modify
+ * the shutdown state should use [ShutdownState] instead of this.
+ */
+@OptIn(ExperimentalApi::class)
+public class MutableShutdownState : ShutdownState() {
```
fractalwrench: This can optionally be annotated with `@ThreadSafe` as documentation

PR #183, opened 2026-02-20
Configure signing information into envars for release workflow

fractalwrench:
## Goal

Cherry pick of #182


breedx-splk (approved):
Looks like the names match what's in the secrets list, even if they don't match the other projects (android)....but I think it's fine.

PR #216, opened 2026-02-23
Alter how the storage directory is defined

fractalwrench:
## Goal

Alters how the storage directory is defined within `:exporters-persistence`. This follows the [approach of SqlDelight](https://sqldelight.github.io/sqldelight/2.1.0/multiplatform_sqlite/#__tabbed_3_1) where passing in `Context` is delegated to the end-user.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -28,7 +28,7 @@ jobs:
           java-version: 21
 
       - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@f29f5a9d7b09a7c6b29859002d29d24e1674c884 # v5.0.1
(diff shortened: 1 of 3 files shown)


bidetofevil (approved):
Making it defined by the user of the module makes sense, but will there be some sensible default or will all users of the SDK have to pass this in explicitly?

fractalwrench:
@bidetofevil I will expose some functions that provide a sensible default in a future changeset.

Ideally I want to avoid passing in `Context` anywhere into opentelemetry-kotlin's API, which is problematic on Android as `Context` is required to obtain the cache directory. I think the solution will be to provide a platform-specific function/class for this processor, which feels reasonable as it's the only place that `Context` is required right now.

PR #217, opened 2026-02-23
Add ShutdownState abstract class and MutableShutdownState

bidetofevil:
-`ShutdownState` abstract class in `sdk-api` with `ifActiveOrElse`, `ifActive`, and `execute` helpers to act based on whether the state is shutdown
- Introduces `MutableShutdownState` in new `sdk-common` module which provides a thread-safe implementation of `ShutdownState` for relevant SDk modules to use.


bidetofevil:
I swore I ran `check` but alas I couldn't have if this stupid error crept in lol

PR #219, opened 2026-02-24
Configure minimum AGP version for library in AarMetadata

fractalwrench:
## Goal

Explicitly configures the minimum AGP version that this library supports [in AarMetadata](https://developer.android.com/reference/tools/gradle-api/8.13/com/android/build/api/variant/AarMetadata#minAgpVersion()). This should be the minimum AGP version supported by the minimum version of Kotlin this library supports (2.0.0 -> AGP 7.1.3).

Doing this makes it more obvious that older AGP versions are unsupported by an improved build-time error message.


fractalwrench:
>But can you actually build with compileSdk 34 and this level of AGP?

Yes! Embrace's SDK actually manages compileSdk 34, which is how I first figured out this was a problem..

PR #223, opened 2026-02-24
Move InstrumentationScopeInfo to sdk-api module

fractalwrench:
## Goal

Moves `InstrumentationScopeInfo` to the `sdk-api` module as it shouldn't be accessible to instrumentation authors.


breedx-splk (approved):

breedx-splk:
I agree that this belongs in the sdk. 😎

PR #225, opened 2026-02-24
Rename fields on `SdkFactory`

fractalwrench:
## Goal

Renames the fields on `SdkFactory` so that they're more idiomatic for Kotlin.

## Testing


api/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/SdkFactory.kt
```diff
@@ -11,30 +11,30 @@ public interface SdkFactory {
...
      * Factory that constructs Span objects.
      */
-    public val spanFactory: SpanFactory
+    public val span: SpanFactory
 
     /**
      * Factory that constructs tracing IDs.
      */
-    public val tracingIdFactory: TracingIdFactory
+    public val identifier: TracingIdFactory
```
breedx-splk: nit: I think I like `tracingIds` (plural) slightly better than the too-vague `identifier`

breedx-splk: alternate idea -- rather than it be plural, split it into two separate interfaces: `SpanIdFactory` and `TraceIdFactory`. 🤷🏻

compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/logging/LogExportTest.kt
```diff
@@ -91,7 +91,7 @@ internal class LogExportTest {
         harness.config.logRecordProcessors.add(contextCapturingProcessor)
 
         // Create a context key and add a test value
-        val currentContext = harness.kotlinApi.contextFactory.implicitContext()
+        val currentContext = harness.kotlinApi.context.implicit()
```
breedx-splk: 👍🏻 nice

breedx-splk (approved):
Renames look good. I offered one suggestion about the `identifier` name, but it wouldn't need to hold up this PR.

PR #226, opened 2026-02-24
Junit6 upgrade attempt

breedx-splk:
Attempt at improving #23 but I have low expectations....


breedx-splk:
> LGTM as a solution for upgrading.
> 
> Longer-term I think the testing module was primarily added with the aim of making it easier for opentelemetry-android to upgrade in future by exposing test rules. I'm not convinced that this is required _right now_, and we're not using it within the project, but I guess we can keep it in if everything compiles...

I'm not opposed to removing it and redoing this when it's needed. I think that's just fine.

PR #227, opened 2026-02-26
Add API stability docs

fractalwrench:
## Goal

Adds some documentation on the stability of the API (or lack thereof). This clarifies how to disable warnings on a project-level basis if folks wish to use the [`optIn` property](https://kotlinlang.org/docs/gradle-compiler-options.html#common-attributes).


bidetofevil:
We should talk about API stability criteria in the next SIG meeting. I think this annotation offers us another level of insulation for things that we are truly not certain about, but for basic things like the shapes of Tracer or Span that we more or less agree on, we can remove the usage of this and let the existing versioning and stability statement guide folks.

PR #235, opened 2026-03-03
Reduce direct use of `SdkFactory`

fractalwrench:
## Goal

Reduces the direct use of `SdkFactory` in preparation for moving its properties to the `OpenTelemetry` interface.

## Testing


breedx-splk (approved):
LGTM!

PR #239, opened 2026-03-05
Remove `create()` function from `TraceFlagsFactory`

fractalwrench:
## Goal

Removes the `create()` API from `TraceFlagsFactory`. This is not strictly required from the OTel spec and is also brittle - if the W3C TraceFlags specification ever adds another bit, we'll need to deprecate this API.

## Testing


bidetofevil (approved):
LGTM as it doesn't seem like we're using it. What was the original use case for adding it?

fractalwrench:
I think this was simply added as syntactic sugar & because opentelemetry-java exposes something similar.

PR #244, opened 2026-03-06
Add shutdown compliance to batch processors

bidetofevil:
## Goal

Use `MutableShutdownState` in the batch processors to store and respect the component's shutdown state. PRs after this will be similar, where stateful processors and exporters will have an internal copy of `MutableShutdownState` and call the appropriate methods to gate method invocations that should respect the state.

Only look at commit `defecf0a893c0a2ad5f9800e69d677c113ca247e` for the change - the other commits are for #217


fractalwrench (approved):
LGTM

PR #248, opened 2026-03-06
Remove deprecated symbols

fractalwrench:
## Goal

Removes symbols that were previously deprecated given that we have no constraints around API compatibility at this time and there has been quite a bit of API churn.


breedx-splk (approved):
LGTM

PR #139, opened 2026-02-11
Regenerate semantic conventions

fractalwrench:
## Goal

Updates the semantic conventions to https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.39.0 by running `./gradlew generateSemanticConventions` to regenerate the Kotlin code.

PR #298, opened 2026-03-13
Add workflow to regenerate semantic conventions

fractalwrench:
## Goal

Attempts to address #139 by adding a workflow to CI that regenerates the semantic conventions.

I'll need to test this out by triggering a workflow, which requires merging to `main` first. There may be a few rounds of iteration.


Msksgm (approved):
LGTM

PR #310, opened 2026-03-16
Replace DelegatingTelemetryCloseable with CompositeTelemetryCloseable

bidetofevil:
`CompositeTelemetryCloseable` does everything `DelegatingTelemetryCloseable` does other than keep state, which is actually good. Move some stuff around so the former can replace the latter


Msksgm (approved):
LGTM

fractalwrench (approved):
LGTM

PR #314, opened 2026-03-17
Hookup AlwaysOn/AlwaysOff samplers via DSL

fractalwrench:
## Goal

Alters the DSL so that the `AlwaysOn/AlwaysOff` samplers are configurable.

## Testing

Added unit tests.


Msksgm (approved):
LGTM

PR #319, opened 2026-03-19
Remove unnecessary annotation

fractalwrench:
## Goal

Kotlin 2.3.20 added a build warning when a DslMarker annotation is added to a use-site where it's a [no-op](https://youtrack.jetbrains.com/projects/KT/issues/KT-81567/Add-a-use-site-warning-if-a-DslMarker-marked-annotation-is-used-on-entities-where-it-is-a-no-op). This was failing the build on #301. I've fixed it by removing the annotation.


bidetofevil (approved):
LGTM

PR #350, opened 2026-03-31
Add API for retrieving current span

### LINKED ISSUE #332: Provide API for getting the currently active Span
## Goal

The API should provide a way to obtain the currently active Span from the current Context, and return a no-op Span if none exists, according to [the spec](https://opentelemetry.io/docs/specs/otel/trace/api/#context-interaction). 

`Span.current()` is the opentelemetry-java approach to this. We should probably add something to `SpanFactory` or `ContextFactory` that achieves the same.

fractalwrench:
## Goal

Adds an API for retrieving the current span to `ContextFactory`. This improves on the current situation where it's necessary to retrieve the `Context` via `implicit()` and then obtain a span from there. This closes #332

## Testing

Added unit tests.


fractalwrench:
Closing in favour of #482

PR #387, opened 2026-04-09
Add initial interface for TextMapPropagator

### LINKED ISSUE #372: Implement TextMapPropagator
Implement TextMapPropagator as defined here: https://opentelemetry.io/docs/specs/otel/context/api-propagators/

fractalwrench:
## Goal

Adds an initial interface for `TextMapPropagator`. This partially resolves #372 but no immediate implementation is planned yet.


bidetofevil (approved):
LGTM. But is there a particular reason you chose to use C instead of T for the type as per normal convention?

PR #392, opened 2026-04-10
Create interface for composite propagators

### LINKED ISSUE #371: Create Propagation interfaces
Create Propagation API interfaces as defined here: https://opentelemetry.io/docs/specs/otel/context/api-propagators/

fractalwrench:
## Goal

Creates an interface for composite propagators. Closes #371


bidetofevil (approved):
LGTM. If we decide that the no-op implementation has to do actually to propagation as well, we can revisit this.

PR #398, opened 2026-04-13
feat(sampling): implement probability sampler

### LINKED ISSUE #336: Implement ProbabilitySampler
Implement ProbabilitySampler as defined in the specification: https://opentelemetry.io/docs/specs/otel/trace/sdk/#probabilitysampler

wpessers:
## Goal

Resolves #336

## Testing

Added unit tests for the relevant code paths.

## Description

I think it's mostly done, I had just a couple of remaining thoughts:
1. Do we have to validate that the `rv` and `th` values are all lowercase, or can we simply assume that when these are written they will adhere to spec? https://opentelemetry.io/docs/specs/otel/trace/tracestate-handling/#pre-defined-opentelemetry-sub-keys states that `th` must have between 1 and 14 lowercase hexadecimal digits and that `rv` must have exactly 14 lowercase hexadecimal digits. In [OtelTraceState.kt#L45-L54](https://github.com/wpessers/opentelemetry-kotlin/blob/a908529d7ffa5af1b6edaed3343e2b8db0cc743f/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/OtelTraceState.kt#L45-L54) we do that validation when getting those values. Should we also include a lowercase check there?
2. Currently the SDK never marks its own randomly generated root trace IDs as Random. `IdGeneratorImpl` generates random IDs: https://github.com/open-telemetry/opentelemetry-kotlin/blob/12a5afd18582f7eac82f639bd239fe04535d5bfe/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/IdGeneratorImpl.kt#L29-L34 But `TraceFlagsFactoryImpl` and `TracerImpl` create span contexts with `isRandom=false`. See: https://github.com/open-telemetry/opentelemetry-kotlin/blob/12a5afd18582f7eac82f639bd239fe04535d5bfe/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/TraceFlagsFactoryImpl.kt#L8-L10 https://github.com/open-telemetry/opentelemetry-kotlin/blob/12a5afd18582f7eac82f639bd239fe04535d5bfe/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt#L102-L114


fractalwrench:
Thanks for contributing this @wpessers. For 1, I think it would be sensible to validate the values are lowercase to be defensive. For 2, that looks like a bug with the current implementation.

I plan on taking a look at this either later today or tomorrow.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/OtelTraceState.kt
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    private fun String.randomness(): Long? =
+        // TODO: is isValidHex() adequate? spec states should be all lowercase
+        takeIf { length == 14 && isValidHex() }
+            ?.toLong(16)
+
+    private fun String.threshold(): Long? =
+        // TODO: is isValidHex() adequate? spec states should be all lowercase
```
fractalwrench: IMO it'd be worth ensuring that the values are lowercase if that's what the spec says.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/ProbabilitySampler.kt
```diff
@@ -0,0 +1,82 @@
...
+        spanKind: SpanKind,
+        attributes: AttributeContainer,
+        links: List<SpanLink>
+    ): SamplingResult {
+        val parentSpanContext = spanFactory.fromContext(context).spanContext
+        val traceState = parentSpanContext.traceState
+        val otelTraceState = OtelTraceState.parse(traceState.get("ot"))
+        otelTraceState.setThreshold(max(otelTraceState.th ?: 0L, rejectionThreshold))
+
+        val explicitRandomness = otelTraceState.rv
```
fractalwrench, line 46: I don't entirely follow the 'explicit randomness' concept [defined here](https://opentelemetry.io/docs/specs/otel/trace/tracestate-handling/#explicit-randomness-value-rv) and how it relates to marking root traces [as random](https://github.com/open-telemetry/opentelemetry-kotlin/blob/12a5afd18582f7eac82f639bd239fe04535d5bfe/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt#L102-L114) as you pointed out in the description. That may be something we can extract into a separate issue to avoid muddying the water...

fractalwrench (commented):
Thanks for taking a stab at this. The approach looked good bar a couple of discussion items you'd already flagged. Once they're resolved I'm happy to approve - I'd also ask other @open-telemetry/kotlin-approvers to review too, given that `ProbabilitySampler` is quite a complex beast.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/OtelTraceState.kt
wpessers: Agree, wanted to get a second opinion before going through. Will update!

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/ProbabilitySampler.kt
wpessers: Agree here also, the `isRandom` "gap" should indeed be tackled in a separate issue.

wpessers:
I had forgotten about this final item on my todo-list, but there is some guidance on how to implement variable precision for the threshold encoding at https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/#converting-floating-point-probability-to-threshold-value
I read that more as a "nice-to-have" not a must-have to be spec-compliant. But I think it's important we implement this at **some** point. Would you want me to open a separate PR or am I good to tack it onto this one as well @fractalwrench?

fractalwrench:
@wpessers I agree with your read of the spec that the requirement is a 'nice-to-have'. I'd suggest opening a separate PR, or an issue to track the work if you don't have time to implement this right now.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/OtelTraceState.kt
```diff
@@ -0,0 +1,57 @@
+package io.opentelemetry.kotlin.tracing.sampling
+
+import io.opentelemetry.kotlin.factory.isValidLowercaseHex
+
+internal class OtelTraceState private constructor(
```
bidetofevil, line 5: Is there a requirement for this object to be threadsafe?

fractalwrench: `Sampler#shouldSample()` must be thread-safe. It achieves this by confining instantiation and use of `OtelTraceState` to a single function so I don't see a thread-safety issue as long as that doesn't change:
https://opentelemetry.io/docs/specs/otel/trace/sdk/#concurrency-requirements

fractalwrench (approved):
LGTM

bidetofevil:
LGTM

bidetofevil (approved):

PR #400, opened 2026-04-14
Tweak pre-release workflow

fractalwrench:
## Goal

Alters the pre-release workflow so that otelbot is configured earlier to see if this solves the branch protection problem seen here: https://github.com/open-telemetry/opentelemetry-kotlin/actions/runs/24236874163


bidetofevil (approved):
LGTM

Issue #47, opened 2026-01-15
Implement environment variable configuration


fractalwrench:
Allow setting the following values from environment variables: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/

fractalwrench:
Closing as this is covered by more specific tickets now

PR #407, opened 2026-04-18
Environment Variable Processing and LogLimit configuration generation

viorel-alexandrescu:
## Goal

The implementation of a feature which will enable OpenTelemetry (KMP version) to be configured via environment variables
To start off, we've chosen to implement the parsing of environment variables configuring LogRecord Limits. 
As per our discussion on #47 

## Testing
Only unit tests for now.


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/OpenTelemetryEnvVarConfigProcessorImpl.kt
```diff
@@ -0,0 +1,21 @@
...
+    private val logLimitProcessor: LogLimitEnvVarConfigProcessor = LogLimitEnvVarConfigProcessorImpl()
+) : OpenTelemetryEnvVarConfigProcessor {
+    override fun process(): EnvironmentConfiguration {
+        return EnvironmentConfiguration(
+            logLimitConfig = logLimitProcessor.resolve { getEnvVarValue(it) }
+        )
+    }
+}
+
+internal expect fun getEnvVarValue(envVar: String): String?
```
fractalwrench, line 23: no-op implementations will be required for non-JVM platforms but this approach looks good

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessor.kt
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Configures LogRecord Limits based on env vars
+ */
+internal sealed class LogLimitEnvVarConfigProcessor :
+    EnvVarConfigProcessor<LogLimitConfig, Int>() {
+    /**
+     * specific list of env vars for log record limits
+     */
+    override val envVars: List<EnvVarName>
+        get() = listOf(
```
fractalwrench: It'd be desirable to only get the envar once when the SDK is initializing so that configuration doesn't become mutable state for the lifetime of a process.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessorImpl.kt
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Configures LogRecord Limits based on env vars
+ */
+internal class LogLimitEnvVarConfigProcessorImpl : LogLimitEnvVarConfigProcessor() {
+    override fun parse(value: String?): Int? = value?.toInt()
+
+    override fun process(entries: Map<EnvVarName, EnvironmentVariable<Int>>): LogLimitConfig {
+        return LogLimitConfig(
+            attributeCountLimit = entries[envVarName("OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT")]?.value
+                ?: DEFAULT_ATTRIBUTE_LIMIT,
```
fractalwrench: Rather than referencing default values in two places (here and also in `LogLimitsConfigImpl`) I'd consider accepting a `defaultValue: LogLimitConfig` parameter in this function. The `LogLimitConfig` generated via `OpenTelemetryConfigDsl` could be passed in.

That would avoid the two default values getting out of sync, and also handles precedence rules for envars vs the DSL.

fractalwrench (commented):
Thanks for drafting this! I left a few initial comments inline about the approach. Could you also sign the CLA?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/OpenTelemetryEnvVarConfigProcessorImpl.kt
viorel-alexandrescu: Not gonna lie, I took inspiration from [this kotlin-env-var library](https://github.com/ScottPierce/kotlin-env-var), but I thought since there aren't many external dependencies in the library, I should implement this mechanism.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessor.kt
viorel-alexandrescu: I see, but how would I hook onto that part of the process? I think these classes would be used only in some internal wiring for envar configuration.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessorImpl.kt
viorel-alexandrescu: Fair point! Thanks for the shout on this, I'll get right to it.

viorel-alexandrescu:
> Thanks for drafting this! I left a few initial comments inline about the approach. Could you also sign the CLA?

With pleasure! I do hope I'm on the right track, I'm working on the comments you left as we speak.
I did sign the CLA, not really sure what's wrong with that check 🤔

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessor.kt
fractalwrench: Basically I think it'd be fine to remove the `get() =` for this property, and have it initialized at object instantiation, given that this class should only get instantiated when we want to load environment variable configuration.

fractalwrench:
@viorel-alexandrescu it looks like CLA is failing on a commit from dependabot for some reason. Given this PR is still in draft, I'd suggest rebasing against `main` and condensing everything down to one commit, as I imagine that'll solve the problem.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessor.kt
viorel-alexandrescu: Fair point, thanks for clarifying 😄

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessor.kt
```diff
@@ -0,0 +1,11 @@
+package io.opentelemetry.kotlin.config.envar.logging
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.config.envar.processor.EnvVarConfigProcessor
+import io.opentelemetry.kotlin.init.config.LogLimitConfig
+
+/**
+ * Configures LogRecord Limits based on env vars
+ */
+@ExperimentalApi
```
fractalwrench: This isn't necessary as the class is `internal` and won't form part of the public API. If it's necessary to suppress a warning then `@OptIn(ExperimentalApi::class)` would be preferable

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessor.kt
```diff
@@ -0,0 +1,11 @@
...
+internal sealed class LogLimitEnvVarConfigProcessor : EnvVarConfigProcessor<LogLimitConfig, Int>()
```
fractalwrench: Does this need to be sealed?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessorImpl.kt
fractalwrench: The implementation doesn't look quite right, although it's part of the way there. I imagined something like this instead:

```
override fun process(
        entries: Map<EnvVarName, EnvironmentVariable<Int>>,
        defaultValue: LogLimitConfig
    ): LogLimitConfig {
        return LogLimitConfig(
            attributeCountLimit = entries[envVarName("OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT")]?.value ?: defaultValue.attributeCountLimit,
...
)
```

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/processor/EnvVarConfigProcessor.kt
```diff
@@ -0,0 +1,33 @@
+package io.opentelemetry.kotlin.config.envar.processor
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.config.envar.model.EnvVarName
+import io.opentelemetry.kotlin.config.envar.model.EnvironmentVariable
+
+
+@ExperimentalApi
```
fractalwrench: Unnecessary as this isn't exposed publicly

fractalwrench (commented):
Thanks for doing this. I've left a few comments inline but think this is nearly there.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessor.kt
viorel-alexandrescu: Ok, thanks for the tip

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/logging/LogLimitEnvVarConfigProcessor.kt
viorel-alexandrescu: Not really I guess, my assumption was to keep these classes as least accessible as possible until otherwise.

fractalwrench (commented):
@viorel-alexandrescu CI is failing due to detekt. It should just be a case of running `./gradlew build` locally and fixing any resulting errors.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/model/EnvVarName.kt
```diff
@@ -0,0 +1,14 @@
+package io.opentelemetry.kotlin.config.envar.model
+
+import kotlin.jvm.JvmInline
+
+@JvmInline
+internal value class EnvVarName(val value: String) {
```
bidetofevil, line 6: Is there a reason why simply using String for the name isn't sufficient and we need to create an abstraction for it?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/processor/EnvVarConfigProcessor.kt
```diff
@@ -0,0 +1,30 @@
+package io.opentelemetry.kotlin.config.envar.processor
+
+import io.opentelemetry.kotlin.config.envar.model.EnvVarName
+import io.opentelemetry.kotlin.config.envar.model.EnvironmentVariable
+
+internal abstract class EnvVarConfigProcessor<R, T> {
```
bidetofevil, line 13: Why is the parsed raw type different than the default type?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/OpenTelemetryEnvVarConfigProcessorImplEntrypoint.kt
```diff
@@ -0,0 +1,44 @@
...
+    )
+    return OpenTelemetryEnvVarConfigProcessorImpl(
+        loggingConfig = cfg.generateLoggingConfig(),
+        logLimitProcessor = logLimitProcessor
+    )
+}
+
+
+internal fun logLimitEnvars() = listOf(
+    envVarName("OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT"),
```
bidetofevil: Does it make sense to define these in a place where they are exposed and can be consumed by implementations like `LogLimitEnvVarConfigProcessorImpl`? Feels like these are a canonically supported set from the API perspective, so folks can see what they are from the code in a single place.

bidetofevil (commented):
Thanks again for your contribution, @viorel-alexandrescu! I have a couple of questions about typing and whether we can simplify them and define the supported ones in a single place, but this looks good in general.

fractalwrench:
After discussion from yesterday's SIG meeting, I've created a milestone that breaks down tasks into more granular issues: https://github.com/open-telemetry/opentelemetry-kotlin/milestone/12

This PR is good from my POV once Hanson's comments are resolved and CI is passing. Thanks for your work so far!

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/model/EnvVarName.kt
viorel-alexandrescu: I thought it would be a good idea because I encapsulated some domain logic in there, ensuring we parse only relevant data.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/processor/EnvVarConfigProcessor.kt
viorel-alexandrescu: What do you mean?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/OpenTelemetryEnvVarConfigProcessorImplEntrypoint.kt
viorel-alexandrescu: Not really sure what to say about this, @fractalwrench's take on this was to extract the list and inject it when building the processor.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/model/EnvVarName.kt
fractalwrench: Feels reasonable to me given this and also that it's a value class, so compiles down to `String` anyway

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/OpenTelemetryEnvVarConfigProcessorImplEntrypoint.kt
fractalwrench: I don't think it should block this PR, but I think Hanson is suggesting defining the constants in one place logically organised so that it's easy to see what constants are supported. E.g.

```
object EnvarConstants {
     object LogLimits {
         const val ATTR_COUNT_LIMIT = envarName("OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT")
         ...
     }
     object SpanLimits
     ...
}
```

viorel-alexandrescu: Hmm, okay, that's a fair shout! I'll amend this.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/processor/EnvVarConfigProcessor.kt
bidetofevil: Reading the code initially, I thought R was the type for the value used if the parsing failed. But looking at how it is used, it is the default used for the return value of the whole processor, while T is the type for each env var's parsed value. Maybe it was the `defaultValue` variable name that tripped me up, or the functions `parse` and `process` being so similar sounding, yet the former is about turning a specific environment variable into a value that is part of the R, while the former is about generating R itself.

Not blocking, but the naming of the code makes it harder to understanding to my mind, so it might be worthwhile to think about whether you can use more precise naming, rather than just `parse`, `process`, and `resolve`, which mean similar things semantically?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/model/EnvVarName.kt
bidetofevil: Fair enough, if you want to do that kind of validation with the name. Is that part of the spec or something we're requiring?

bidetofevil (approved):
My comments are non-blocking, and the core of this looks sound. Thanks for the contribution!

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/model/EnvVarName.kt
viorel-alexandrescu: I don't think it's a requirement, I thought at the time it would be a good guard against error cases.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/processor/EnvVarConfigProcessor.kt
viorel-alexandrescu: That's a fair shout, I'll have a go at refactoring this.

viorel-alexandrescu: I've pushed some changes to help with your notice, but I think we should come back to this in the future.
I'll resolve the conversation for now.

fractalwrench:
@viorel-alexandrescu CI is failing due to detekt. Once that's resolved I'll approve & merge.

fractalwrench (approved):
LGTM - thanks for your contribution!

PR #412, opened 2026-04-21
Revert workflow changes

fractalwrench:
## Goal

Reverts workflow changes made in #400 so that we use the same workflow approach as opentelemetry-android.


Msksgm (approved):
LGTM

bidetofevil (approved):
LGTM

PR #424, opened 2026-04-22
Prevent empty key for attributes

### LINKED ISSUE #419: Attributes should not allow empty strings as key values
## Goal

Attribute keys should be a non-empty string. Currently we don't validate against non-empty strings for keys: https://opentelemetry.io/docs/specs/otel/common/#attribute

fractalwrench:
## Goal

Closes #419 by preventing empty keys from being used in `AttributesModel`.


bidetofevil (approved):
LGTM

PR #429, opened 2026-04-23
Ensure that resource attributes don't have limits

### LINKED ISSUE #420: Resource should be exempt from Attributes limit
The spec states that `Resource` should be exempt from the attributes limit but the default attribute limit is currently applied in `ResourceImpl`: https://opentelemetry.io/docs/specs/otel/common/#exempt-entities

fractalwrench:
## Goal

Ensures that resource attributes are exempt from limits. Closes #420.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #430, opened 2026-04-23
Support ByteArray in attributes

### LINKED ISSUE #422: Support ByteArray value in Attributes
Attributes should support a ByteArray value type: https://opentelemetry.io/docs/specs/otel/common/#attribute

fractalwrench:
## Goal

Supports ByteArray as a type that can be added to Attributes. Closes #422.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. I guess we are getting ahead of Java with this, which means using the compat impl will make the SDK miss some functionality, eh? Another reason to upgrade I guess

PR #431, opened 2026-04-23
Allow a custom HttpClient for OTLP export

### LINKED ISSUE #416: Allow passing custom Ktor client in OTLP exporters
Originally discussed in #409. It would be nice to have the ability to specify a custom Ktor client used for OTLP export (and any other exporters that initiate http requests).

>We need to use our own ktor client to upload the traces and spans, with a slightly different URL, protected with access token and client certificate and it handles token renewals automatically
Also, the ktor client we have to use is also the one that is instrumented, so we need to filter out the traces/spans that are caused by flushing the spans and logs, to avoid an infinite loop

fractalwrench:
## Goal

Allows a custom HttpClient for OTLP export. Closes #416.


bidetofevil (approved):
Nice. This means folks can provide a different, uninstrumented client so they don't have to worry about excluding the OTLP export path when logging HTTP requests.

PR #432, opened 2026-04-23
Handle unexpected types in attributes

### LINKED ISSUE #409: Support integers in AttributesProtobufConverter
When using the protobuf exporter directly via `List<SpanData>.toProtobufByteArray()` it throws an unsupported operation when I fill in my SpanData's attributes with an integer instead of a long.

Would it make sense to support Integers as well?

Fixing it is as easy as adding a case for integers in the AttributesProtobufConverter, and either sending them as integers (if the protocol allows), or converting them to longs under the hood.

https://github.com/open-telemetry/opentelemetry-kotlin/blob/12a5afd18582f7eac82f639bd239fe04535d5bfe/exporters-protobuf/src/commonMain/kotlin/io/opentelemetry/kotlin/export/conversion/AttributesProtobufConversion.kt#L35

I am not sure if this breaks expectations when converting the bytearray back into a list of SpanDatas via `ByteArray.toSpanDataList()`, because it can break round-trip consistency if you silently convert integers to longs.

fractalwrench:
## Goal

Handles unexpected types in attributes by casting them to either `Long` or `Double`. Closes #409.

## Testing

Added unit tests.

### DIFF
--- a/exporters-protobuf/src/commonMain/kotlin/io/opentelemetry/kotlin/export/conversion/AttributesProtobufConversion.kt
+++ b/exporters-protobuf/src/commonMain/kotlin/io/opentelemetry/kotlin/export/conversion/AttributesProtobufConversion.kt
@@ -34,6 +34,8 @@ private fun convertAttributeValue(value: Any): AnyValue = when (value) {
     is String -> AnyValue(string_value = value)
     is Long -> AnyValue(int_value = value)
     is Double -> AnyValue(double_value = value)
+    is Float -> AnyValue(double_value = value.toDouble())
+    is Number -> AnyValue(int_value = value.toLong())
     is Boolean -> AnyValue(bool_value = value)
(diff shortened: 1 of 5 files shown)


exporters-protobuf/src/commonMain/kotlin/io/opentelemetry/kotlin/export/conversion/AttributesProtobufConversion.kt
```diff
@@ -34,6 +34,8 @@ private fun convertAttributeValue(value: Any): AnyValue = when (value) {
     is String -> AnyValue(string_value = value)
     is Long -> AnyValue(int_value = value)
     is Double -> AnyValue(double_value = value)
+    is Float -> AnyValue(double_value = value.toDouble())
+    is Number -> AnyValue(int_value = value.toLong())
```
bidetofevil, line 38: Ah nice

bidetofevil (approved):
LGTM

PR #436, opened 2026-04-24
Implement baggage API

fractalwrench:
## Goal

Provides an implementation of `Baggage` in the implementation module.


bidetofevil (approved):
LGTM

Issue #376, opened 2026-04-07
Create MeterProvider API


fractalwrench:
Create a `MeterProvider` interface and expose it on the `OpenTelemetry` entry point: https://opentelemetry.io/docs/specs/otel/metrics/api/

inetand:
Hi @fractalwrench — I’m currently working on this and have a clarification question.

Do I understand correctly that exposing MeterProvider via the OpenTelemetry entry points requires adding an implementation as well?

If so, I’m planning to split this into multiple PRs (DSL, implementation, and entry points). Does it make sense?

fractalwrench:
@inetand that makes sense to me.

PR #437, opened 2026-04-25
Add MeterProvider API

inetand:
## Goal

Adds the `MeterProvider` interface. Wiring with the OpenTelemetry entry point will be added in a subsequent PR.
Partially addresses https://github.com/open-telemetry/opentelemetry-kotlin/issues/376.

## Testing

Since this PR only adds the API surface, no behavioral impact is expected. Build and detekt checks pass.


fractalwrench (approved):
LGTM - thanks!

Issue #440, opened 2026-04-28
Support AnyValue


fractalwrench:
Add support for AnyValue as an attribute type: https://opentelemetry.io/docs/specs/otel/common/#attribute

PR #453, opened 2026-04-28
Provide implementation of AnyValue

fractalwrench:
## Goal

Implements AnyValue as in the [OpenTelemetry specification](https://opentelemetry.io/docs/specs/otel/common/#anyvalue). I've chosen a sealed inheritance hierarchy as this allows exhaustive `when` over each type, at the cost of introducing concrete classes into the `api` module.

## Testing

Added unit tests.

### DIFF
--- a/api/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AnyValue.kt
+++ b/api/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AnyValue.kt
@@ -10,6 +10,10 @@ import io.opentelemetry.kotlin.ExperimentalApi
 @ExperimentalApi
 public sealed class AnyValue {
 
+    public object NullValue : AnyValue() {
+        override fun toString(): String = "NullValue"
+    }
+
     public class StringValue(public val value: String) : AnyValue() {
         override fun equals(other: Any?): Boolean {
             if (this === other) {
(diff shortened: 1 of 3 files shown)


bidetofevil (approved):
Ironically, `AnyValue` isn't `Any` value lol

LGTM

api/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AnyValue.kt
```diff
@@ -0,0 +1,124 @@
...
+            }
+            return value.contentEquals(other.value)
+        }
+
+        override fun hashCode(): Int = value.contentHashCode()
+
+        override fun toString(): String = "BytesValue(size=${value.size})"
+    }
+
+    public class ArrayValue(public val values: List<AnyValue>) : AnyValue() {
```
bidetofevil: Are there any concerns that we are using List and Array interchangeably in this context even though they are different things in Kotlin? e.g. perf implications when a not-indexed List is passed in and the initialization expects otherwise

bidetofevil (approved):
LGTM. One non-blocking comment about the use of Array and List interchangeably in verbiage, but are different things in Kotlin.

breedx-splk (approved):
This looks good to me...but I think we should consider adding (in this PR) the null object case. I know it's often expected that the usage will be of a literal null (eg. `AnyValue?` nullable), but sometimes it's nice to avoid that viral nullability baggage. The spec also wants us to have this.

breedx-splk:
This resolves #440.

api/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AnyValue.kt
fractalwrench: Renamed to `ListValue`.

bidetofevil (approved):
LGTM

PR #454, opened 2026-04-28
Attributes extension -- handle int in map and improved js handling

breedx-splk:
There were two cases that seemed to not be working as well as expected with the extension -- Int types in the map and some double/floating values being converted to long by the js backend.

### DIFF
--- a/api-ext/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesMutatorExt.kt
+++ b/api-ext/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesMutatorExt.kt
@@ -17,16 +17,21 @@ public fun AttributesMutator.setAttributes(attributes: Map<String, Any>) {
         when (val input = it.value) {
             is String -> setStringAttribute(it.key, input)
             is Boolean -> setBooleanAttribute(it.key, input)
-            is Int -> setLongAttribute(it.key, input.toLong())
             is Long -> setLongAttribute(it.key, input)
-            is Double -> setDoubleAttribute(it.key, input)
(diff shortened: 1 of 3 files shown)


api-ext/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesMutatorExt.kt
```diff
@@ -18,9 +18,7 @@ public fun AttributesMutator.setAttributes(attributes: Map<String, Any>) {
             is String -> setStringAttribute(it.key, input)
             is Boolean -> setBooleanAttribute(it.key, input)
             is Long -> setLongAttribute(it.key, input)
```
fractalwrench, line 20: Should `setNumericAttribute` replace this line too?

fractalwrench (commented):
One query about why `setLongAttribute` isn't also replaced by `setNumericAttribute`, otherwise this LGTM

api-ext/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesMutatorExt.kt
breedx-splk: Oh I think it totally can, and in fact I tested that before submitting the PR -- however it avoids the extra instance checks, which I optimistically/naively assumed was more efficient. How do you feel about that?

fractalwrench: I feel great about anything that's more efficient!

fractalwrench (approved):
LGTM

PR #456, opened 2026-04-29
Implement storage of baggage on context

fractalwrench:
## Goal

Implements storage of Baggage on the `Context` object, following the rules in [the spec](https://opentelemetry.io/docs/specs/otel/baggage/api/#context-interaction).


bidetofevil (approved):
LGTM

PR #458, opened 2026-04-30
Implement W3CBaggagePropagator

fractalwrench:
## Goal

Adds an implementation for the W3CBaggagePropagator as defined in the spec: https://opentelemetry.io/docs/specs/otel/context/api-propagators/#propagators-distribution

## Testing

Added unit tests.


bidetofevil (approved):
LGTM from what I can tell. I'm letting the tests do the heavylifting for verifying everything works according to spec 😅

PR #460, opened 2026-04-30
Create compat adapter for TextMapPropagator

fractalwrench:
## Goal

Creates a compatibility adapter for `TextMapPropagator`.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. 

Non-blocking thought: I wonder if it's worth testing that the SDK and compat implementations, for the same set of key-values, produces identical results? This goes for the baggage stuff too. May be a good separate issue that is a good first-task for pontential contributers?

PR #461, opened 2026-04-30
Implement compat layer for PropagatorFactory

fractalwrench:
## Goal

Implements a compat layer for `PropagatorFactory`.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #462, opened 2026-04-30
Add traceparent implementation

fractalwrench:
## Goal

Adds an implementation of [traceparent](https://www.w3.org/TR/trace-context/#traceparent-header). This will be required by the trace context propagator. 

## Testing

Added unit tests.

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceParent.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceParent.kt
@@ -1,0 +1,109 @@
+package io.opentelemetry.kotlin.propagation
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.factory.TraceFlagsFactory
+import io.opentelemetry.kotlin.tracing.TraceFlags
+
+/**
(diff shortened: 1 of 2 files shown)


bidetofevil (approved):
LGTM

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceParent.kt
```diff
@@ -0,0 +1,109 @@
...
+@OptIn(ExperimentalApi::class)
+internal class TraceParent(
+    val version: String,
+    val traceId: String,
+    val spanId: String,
+    val traceFlags: TraceFlags,
+) {
+
+    init {
+        require(version.length == VERSION_LEN && version.isLowerHex()) {
```
bidetofevil, line 21: Side note: I'm beginning to see a lot more of using `require()` to validate assumptions and invariants at instance creation time. I'd rather not default to this pattern, as runtime exceptions can crash the app it's running on, and it feels too big a hammer for something like an instrumentation SDK to wield. Perhaps we can discuss this in the next SIG meeting.

fractalwrench: Discussing at the next SIG sounds good

PR #464, opened 2026-05-01
Implement TraceState

fractalwrench:
## Goal

Following on from #462 create a class that can encode/decode the trace state header.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceParentImpl.kt
```diff
@@ -10,7 +10,7 @@ import io.opentelemetry.kotlin.tracing.TraceFlags
  * https://www.w3.org/TR/trace-context/#traceparent-header
  */
 @OptIn(ExperimentalApi::class)
-internal class TraceParent(
+internal class TraceParentImpl(
```
breedx-splk: Why this name change? I would only expect to find an `Impl` where there is an interface...but this is just a regular old concrete class. I'd prefer to leave it as `TraceParent` unless there's a good reason for the clunky change.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceStateImpl.kt
```diff
@@ -0,0 +1,67 @@
...
+import io.opentelemetry.kotlin.factory.TraceStateFactory
+import io.opentelemetry.kotlin.tracing.TraceState
+
+/**
+ * Implementation of a W3C `tracestate` header.
+ *
+ * https://www.w3.org/TR/trace-context-2/#tracestate-header
+ */
+@OptIn(ExperimentalApi::class)
+internal class TraceStateImpl(traceState: TraceState) {
```
breedx-splk: This class now has the same name as the `TraceStateImpl` in the tracing (sdk) implementation, which I think could be somewhat confusing.

Maybe because this class is primarily concerned with encoding/decoding, maybe we just split it in two and call them `TraceStateEncoder` and `TraceStateDecoder` respectively?

implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/propagation/TraceStateMarshallerTest.kt
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @Test
+    fun `decode silently drops list-members beyond the 32-entry cap`() {
+        val header = (1..33).joinToString(",") { "k$it=v$it" }
+        val expected = (1..32).joinToString(",") { "k$it=v$it" }
+        val ts = TraceStateImpl.decode(header, factory)
+        assertEquals(expected, ts.encode())
+    }
+}
```
breedx-splk, line 160: These tests are great. I'm going to also add an issue to add fuzz tests around this. Not only is that just a good idea that could potentially uncover something unexpected, it helps the security rating.

breedx-splk: Created #471.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceParentImpl.kt
fractalwrench: I've reverted the name change

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceStateImpl.kt
fractalwrench: Renamed as `TraceStateMarshaller`

PR #466, opened 2026-05-01
Implement getAll() on TextMapGetter

fractalwrench:
## Goal

Adds an implementation for `getAll()` on the TextMapGetter interface, as per the spec: https://opentelemetry.io/docs/specs/otel/context/api-propagators/#getall


bidetofevil (approved):
LGTM

PR #468, opened 2026-05-01
Truncate byte array attribute values

fractalwrench:
## Goal

Truncates byte array values in attributes so that they follow the [attribute limits](https://opentelemetry.io/docs/specs/otel/common/#attribute-limits) defined in the spec.

## Testing

Added unit tests.

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesModel.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesModel.kt
@@ -72,7 +72,11 @@ internal class AttributesModel(
 
     override fun setByteArrayAttribute(key: String, value: ByteArray) {
         if (canAddAttribute(key)) {
-            attrs[key] = value
+            attrs[key] = if (value.size > attributeValueLengthLimit) {
+                value.copyOf(attributeValueLengthLimit)
+            } else {
+                value
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesModel.kt
```diff
@@ -72,7 +72,11 @@ internal class AttributesModel(
 
     override fun setByteArrayAttribute(key: String, value: ByteArray) {
         if (canAddAttribute(key)) {
-            attrs[key] = value
+            attrs[key] = if (value.size > attributeValueLengthLimit) {
```
bidetofevil, line 75: Do want to to drop this entirely (or put empty string in) rather than creaeting a corrupted bytestring without indication that we are doing so?

fractalwrench: The requirement from the spec is:

>if it is a byte array, if its length exceeds that limit (counting each byte as 1), SDKs MUST truncate that value, so that its length is at most equal to the limit,

https://opentelemetry.io/docs/specs/otel/common/#attribute-limits

bidetofevil: Truncating is odd. Is it even usable after? If that's what the spec says, I guess.

bidetofevil: >  so that its length is at most equal to the limit

Technically, truncating to 0 bytes also fits the letter of the spec. Do we really want to send that much data that will likely not be usable?

Anyway, I'm OK with this, but this doesn't seem that pragmatic.

bidetofevil (approved):

PR #469, opened 2026-05-01
Create entrypoint for propagators

fractalwrench:
## Goal

Creates an entrypoint for retrieving a `TextMapPropagator` from the `OpenTelemetry` interface, which forms part of the instrumentation API. The SDK can be configured to allow custom propagators using a new DSL that is added in this changeset.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #470, opened 2026-05-01
Allow supplying a custom implicit context storage mechanism

fractalwrench:
## Goal

Allows users instantiating the SDK to supply a custom storage mechanism for implicit context.

## Testing

Added unit tests.


bidetofevil (approved):
Nice! Have we settled on what we are defaulting to now?

fractalwrench:
I think that's still up for discussion.

PR #475, opened 2026-05-04
Add the gradle wrapper checksum

breedx-splk:
We have this nice [gradle wrapper workflow](https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/.github/workflows/gradle-wrapper-validation.yml), but I think it requires the checksum to be present to really be meaningful. We have this in other repos, and renovate will update it.


fractalwrench (approved):
LGTM

PR #479, opened 2026-05-05
Implement ThreadLocal implicit context storage

### LINKED ISSUE #477: Create thread based context storage
Creates a thread-based approach for storing the implicit context

fractalwrench:
## Goal

Implements a storage solution for implicit context that uses `ThreadLocal`. This is _not_ made the default behavior in this changeset and is included to gain parity with opentelemetry-java's feature set. A future changeset will attempt to add a coroutine-based implicit context storage which will arguably be more suitable for a KMP library. However, no solution will get automatic context 100% right in all cases, so I feel it's sensible to offer multiple solutions and allows library consumers to pick the most appropriate for their needs.

Closes #477.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. Offering as an alternative is a good way to satisfy backwards compat. Not being the default is really my big hang up.

PR #482, opened 2026-05-05
Alter how spans are stored on Context

### LINKED ISSUE #332: Provide API for getting the currently active Span

### LINKED ISSUE #476: Provide helper for extracting Baggage from Context
Provide a helper for extracting `Baggage` from the implicit `Context`: https://opentelemetry.io/docs/specs/otel/baggage/api/#context-interaction

fractalwrench:
## Goal

This changeset alters how span objects are stored and extracted from `Context`. Previously the syntax for extracting a span would look like this:

```
val ctx = otel.context.makeCurrent(span)
val retrievedSpan = otel.spanFactory.fromContext(otel.context.implicit())
```

Whereas after the changes the API looks like this:

```
val root = otel.context.root()
val ctx = root.storeSpan(span)
val retrievedSpan = root.extractSpan()
```

IMO this is preferable to the previous API as it's more obvious that spans can be stored/retrieved on a `Context` object. Additionally, it matches the API used for `Baggage`.

### DIFF
--- a/api/api/jvm/api.api
+++ b/api/api/jvm/api.api
@@ -57,9 +57,11 @@ public abstract interface class io/opentelemetry/kotlin/context/Context {
 	public abstract fun attach ()Lio/opentelemetry/kotlin/context/Scope;
 	public abstract fun clearBaggage ()Lio/opentelemetry/kotlin/context/Context;
 	public abstract fun extractBaggage ()Lio/opentelemetry/kotlin/baggage/Baggage;
+	public abstract fun extractSpan ()Lio/opentelemetry/kotlin/tracing/Span;
 	public abstract fun get (Lio/opentelemetry/kotlin/context/ContextKey;)Ljava/lang/Object;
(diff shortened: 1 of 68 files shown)


bidetofevil (approved):
This looks great. Not having to fake `makeCurrent()` and explain why the behavior is so different between the implementations, yet still offering an explicit way to do the thread-local current context in Java, seems like the right thing.

Instrumentation written against the Java API already need to be changed to work with the Kotlin API - this just makes a previously easy to over-look difference more apparent, which is a good thing IMO.

Awesome work!

PR #484, opened 2026-05-06
Add link to getting started instructions

fractalwrench:
## Goal

Adds a link to opentelemetry.io which contains the canonical guide for getting started with the SDK. I've removed redundant content from the README that now lives in the guide.


bidetofevil (approved):
LGTM

PR #492, opened 2026-05-08
Add fuzz tests for trace context

### LINKED ISSUE #471: Add W3C trace context fuzz tests
[Similar to how java does it](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/test/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagatorFuzzTest.java), we should have a set of fuzz tests that always run and exercise the encoding/decoding of w3c trace context to look for unexpected outcomes.

I'm not entirely sure what kotlin frameworks exist for this.

fractalwrench:
## Goal

Closes #471 by adding fuzz tests for W3C trace context, [using Kotest](https://kotest.io/).

### DIFF
--- a/implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/propagation/W3CTraceContextPropagatorFuzzTest.kt
+++ b/implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/propagation/W3CTraceContextPropagatorFuzzTest.kt
@@ -16,6 +16,7 @@ import io.opentelemetry.kotlin.tracing.TraceFlagsImpl
 import kotlinx.coroutines.test.runTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
 /**
@@ -105,6 +106,8 @@ internal class W3CTraceContextPropagatorFuzzTest {
                     extracted.extractSpan().spanContext.isValid,
                     "non-root context with invalid SpanContext from carrier=$carrier",
                 )
+            } else {
+                assertFalse(extracted.extractSpan().spanContext.isValid)
             }
         }
     }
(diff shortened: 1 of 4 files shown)


implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/propagation/W3CTraceContextPropagatorFuzzTest.kt
```diff
@@ -0,0 +1,164 @@
...
+                    put("tracestate", ts)
+                }
+            }
+            val extracted = propagator.extract(root, carrier, MapTextMapGetter)
+            if (extracted != root) {
+                assertTrue(
+                    extracted.extractSpan().spanContext.isValid,
+                    "non-root context with invalid SpanContext from carrier=$carrier",
+                )
+            }
```
breedx-splk, line 111: I get twitchy when I see assertions in loops or in conditionals, because it introduces failure modes in which the assertion never even runs. In this case, if the propagated context wasn't set up correctly, it could always/only be root, and the assertion was never run and the test still passes. That might not actually ever be possible, but to a reader, it's not immediately obvious.

One way to fix that is to have the matching `else` for the `if` and put assertions in that too....or if that doesn't make sense, set a flag inside/after the assertion and verify it at the end of the test.

implementation/src/jvmTest/kotlin/io/opentelemetry/kotlin/propagation/W3CTraceContextPropagatorKFuzzTest.kt
```diff
@@ -0,0 +1,5 @@
+package io.opentelemetry.kotlin.propagation
+
+// Intentionally empty. Previously held a kotlinx.fuzz attempt before discovering its
+// publishing channel (plan-maven.apal-research.com) is unreachable. Fuzz tests now
+// live in commonTest using Kotest property-based testing for KMP coverage.
```
breedx-splk: 👍🏻

breedx-splk (approved):
This is cool and I had no idea that kotest had this. Very nice.

I did leave one comment, but it's not that critical. Thanks for turning this out so quickly!

PR #498, opened 2026-05-11
Scaffold MeterProvider config and compat layer

inetand:
## Goal

Partially addresses #376 by scaffolding `MeterProvider` configuration on the Kotlin implementation and compat layers.

This PR adds:
- `MeterProvider` config DSL + implementation scaffolding
- Java ↔ Kotlin compat adapters for `MeterProvider` / `Meter`
- Java OTel metric typealiases
- Basic compat-layer no-op metric delegation until Kotlin instruments land

Entry-point wiring, metric readers, views, and the aggregation pipeline are intentionally deferred to a subsequent PR.

## Testing

- Added config + adapter unit tests
- Verified `build`, `assembleAndroidTest`, and `apiDump`

### DIFF
--- a/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/init/CompatMeterProviderConfig.kt
+++ b/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/init/CompatMeterProviderConfig.kt
@@ -20,12 +20,9 @@ internal class CompatMeterProviderConfig(
 ) : MeterProviderConfigDsl {
 
     private val builder: OtelJavaSdkMeterProviderBuilder = OtelJavaSdkMeterProvider.builder()
-    private var serviceNameOverride: String? = null
-
-    override var serviceName: String
-        get() = serviceNameOverride ?: "unknown_service"
+    override var serviceName: String = "unknown_service"
         set(value) {
-            serviceNameOverride = value
+            field = value
             resourceAttrs.setStringAttribute(ServiceAttributes.SERVICE_NAME, value)
         }
 
--- a/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/init/CompatMeterProviderConfig.kt
+++ b/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/init/CompatMeterProviderConfig.kt
@@ -1,0 +1,59 @@
+package io.opentelemetry.kotlin.init
+
+import io.opentelemetry.kotlin.Clock
+import io.opentelemetry.kotlin.ExperimentalApi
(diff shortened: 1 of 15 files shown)


compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/metrics/MeterAdapter.kt
```diff
@@ -0,0 +1,9 @@
+package io.opentelemetry.kotlin.metrics
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.aliases.OtelJavaMeter
+
+@ExperimentalApi
+internal class MeterAdapter(
+    @Suppress("UnusedPrivateProperty") private val impl: OtelJavaMeter,
```
inetand, line 8: Not sure about the convention around suppressing unused properties, but I left it intentionally for now. 
I plan to remove this once an implementation is available.

fractalwrench: That feels reasonable to me. Thanks for opening this, ping me if you have any other questions

compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/init/CompatMeterProviderConfig.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    private val builder: OtelJavaSdkMeterProviderBuilder = OtelJavaSdkMeterProvider.builder()
+    private var serviceNameOverride: String? = null
+
+    override var serviceName: String
+        get() = serviceNameOverride ?: "unknown_service"
+        set(value) {
+            serviceNameOverride = value
+            resourceAttrs.setStringAttribute(ServiceAttributes.SERVICE_NAME, value)
+        }
```
fractalwrench:
```suggestion
    override var serviceName: String = "unknown_service"
        set(value) {
            field = value
            resourceAttrs.setStringAttribute(ServiceAttributes.SERVICE_NAME, value)
        }
```

nit: is there a reason this sort of syntax can't be used?

fractalwrench (approved):
LGTM pending one comment. Thanks for taking this on!

compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/init/CompatMeterProviderConfig.kt
inetand: Fair point! I originally wrote this in the same style as the other configs, but backing-field syntax is a perfect fit here indeed. I’ll switch to it, thanks!

Issue #502, opened 2026-05-13
SDK should not throw runtime exceptions that can crash the app


bidetofevil:
As brought up in the review for #462 , sprinkled throughout the SDK are invariant validations that thrown runtime exceptions (e.g. using `require()`) during object initialization that aren't not guaranteed to be caught. Since the telemetry is optional functionality, it should never lead to exceptions at runtime or build due to poor configuration.

We should either throw a checked exception that has to be explicitly handled or fail in a not-crashy way, with the latter being preferable in most cases.

fractalwrench:
Agree, we certainly shouldn't be throwing at runtime after the SDK has initialized. During SDK initialization I can see more of an argument for throwing exceptions if the configuration is useless, but IMO it's not what we want as a default.

It'd probably be worth updating AGENTS.md to include this as a guideline given how critical it is and how easy it is for code to creep in.

PR #503, opened 2026-05-13
Don't throw when using the DSL to configure exporters and processors

bidetofevil:
Instead of throwing when composites aren't initialized with concrete processors and exporters to wrap, add in a no-op implementation and log platform error instead.

I suppose throwing at build time isn't the worst thing and surfaces configuration errors more loudly, but this seems a slight overstep. Interested to hear what y'all think. Once we figure out the parameters, I can submit the other changes that follow a similar pattern (at runtime) where we don't throw but error out in a less destructive way.

Partially addresses #502


fractalwrench (commented):
Can you add test coverage to assert that the noop processor/exporter implementations don't throw? Otherwise I think this looks good to go.

bidetofevil:
> Can you add test coverage to assert that the noop processor/exporter implementations don't throw? Otherwise I think this looks good to go.

Done through testing the empty composite, which is the thing to verify anyway

fractalwrench (approved):
LGTM

PR #504, opened 2026-05-13
Invalid configs should not throw exceptions

bidetofevil:
Report and SDK error and fallback to defaults when processor and exporter configs are not valid

While these are still only currently used in the DSL, you can imagine that they may not be in the future.

Partially addresses #502

### DIFF
--- a/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/export/BatchTelemetryProcessor.kt
+++ b/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/export/BatchTelemetryProcessor.kt
@@ -25,7 +25,7 @@ internal class BatchTelemetryProcessor<T>(
     private val exportAction: suspend (telemetry: List<T>) -> OperationResultCode,
 ) : TelemetryCloseable {
 
-    private val validMaxQueueSize: Int = validateOrUseDefault(
+    private val maxQueueSize: Int = validateOrUseDefault(
         sdkErrorHandler = sdkErrorHandler,
         api = API,
         configParameterName = "maxQueueSize",
@@ -33,7 +33,7 @@ internal class BatchTelemetryProcessor<T>(
         default = BatchTelemetryDefaults.MAX_QUEUE_SIZE,
     ) { it >= 0 }
 
-    private val validScheduleDelayMs: Long = validateOrUseDefault(
+    private val scheduleDelayMs: Long = validateOrUseDefault(
         sdkErrorHandler = sdkErrorHandler,
         api = API,
         configParameterName = "scheduleDelayMs",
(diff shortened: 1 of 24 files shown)


exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/export/BatchTelemetryProcessor.kt
```diff
@@ -25,31 +25,31 @@ internal class BatchTelemetryProcessor<T>(
     private val exportAction: suspend (telemetry: List<T>) -> OperationResultCode,
 ) : TelemetryCloseable {
 
-    private val maxQueueSize: Int = validateOrUseDefault(
+    private val validMaxQueueSize: Int = validateOrUseDefault(
```
fractalwrench: IMO I'd just keep the previous name for these variables (and similar places within other files). `validateOrUseDefault` makes it apparent what is happening here

exporters-core/src/commonTest/kotlin/io/opentelemetry/kotlin/export/BatchTelemetryProcessorTest.kt
```diff
@@ -10,75 +11,24 @@ import kotlinx.coroutines.test.advanceUntilIdle
...
-        assertFailsWith<IllegalArgumentException> {
-            BatchTelemetryProcessor<Unit>(
-                maxQueueSize = 100,
-                scheduleDelayMs = 1,
-                exportTimeoutMs = 1,
-                maxExportBatchSize = 200,
-                exportAction = { OperationResultCode.Success }
-            )
-        }
+    fun testInvalidValuesDoesNotThrow() {
```
fractalwrench: Ideally it'd be good to have coverage of what the defaults are set as. That might require restructuring the code somewhat

exporters-persistence/src/commonMain/kotlin/io/opentelemetry/kotlin/export/PersistedTelemetryConfig.kt
```diff
@@ -1,25 +1,44 @@
 package io.opentelemetry.kotlin.export
 
+import io.opentelemetry.kotlin.config.validateOrUseDefault
+import io.opentelemetry.kotlin.error.NoopSdkErrorHandler
+import io.opentelemetry.kotlin.error.SdkErrorHandler
+
 internal class PersistedTelemetryConfig(
```
fractalwrench, line 7: I think it'd be worth following this pattern for the `BatchTelemetryProcessor` as it'd make it easier to write unit tests asserting the default is used (and also separates the processor from validation logic)

exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/export/BatchTelemetryProcessor.kt
bidetofevil: Sure. This was mostly so we don't use the same variable name for the constructor parameter and the member, but we can always just use `this` to disambiguate.

fractalwrench (approved):
LGTM

PR #506, opened 2026-05-15
Remove throws on init in trace and span context usage scenarios

bidetofevil:
Instead of throwing when invalid value are encountered, return no-op implementations. Use factory method when appropriate and make constructors private. Added some additional test coverage while I was at it.

Partially address #502


fractalwrench (approved):
LGTM

PR #514, opened 2026-05-19
Drop support for iOS x64

### LINKED ISSUE #513: Drop ios x64 (intel) support
It looks like kotlin has deprecated the ios x64 targets ([source](https://kotlinlang.org/docs/whatsnew2220.html#deprecation-of-x86-64-apple-targets)). As a result of this, other libraries (such as jetbrains compose, which we use here) have started dropping support for it as well ([source](https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.11.0-alpha02)).

I think the simplest thing that we can do is stick with the status quo and also drop support. I think it also makes the most sense, because this is a kotlin language project, and if kotlin doesn't support it then it becomes very difficult for us to.

This dependency PR is blocked due to this: https://github.com/open-telemetry/opentelemetry-kotlin/pull/510

The usages here are pretty straghtforward to find in gradle files/config...simply search for `iosx64`.

fractalwrench:
## Goal

Drop support for iOS x64, closes #513.

### DIFF
--- a/buildSrc/src/main/kotlin/io/opentelemetry/kotlin/KotlinConfig.kt
+++ b/buildSrc/src/main/kotlin/io/opentelemetry/kotlin/KotlinConfig.kt
@@ -36,7 +36,6 @@ fun Project.configureKotlin(
             val framework = XCFramework(frameworkName)
 
             listOf(
-                iosX64(),
                 iosArm64(),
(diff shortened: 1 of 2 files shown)


bidetofevil (approved):
Didn't know this was still a thing?

PR #516, opened 2026-05-19
Support setting attribute with AnyValue type

### LINKED ISSUE #499: Support setting attributes of type AnyValue
Right now, we support all the otel primitive and list types, but we don't yet allow setting attributes of type AnyValue in the AttributesMutator. For spec compliance, we think we need to add this.

fractalwrench:
## Goal

Closes #499 by allowing the `AnyValue` type to be added to attributes.


bidetofevil (approved):
LGTM

Issue #515, opened 2026-05-19
Minimum supported Kotlin version is higher than stated in documentation


fractalwrench:
The documentation states that Kotlin 2.0 and greater are supported: https://opentelemetry.io/docs/languages/kotlin/getting-started/#supported-platforms

However, it appears a transitive dependency is pulling in Kotlin stdlib 2.2 which means that library consumers can't use a version less than Kotlin 2.0. We should identify which library is doing this and ideally add some automated coverage to prevent regressions in future.

PR #517, opened 2026-05-19
Downgrade dependencies

fractalwrench:
## Goal

Downgrades several dependencies to where they were when 0.3.0 was released. I've confirmed that this works in an app that compiles with Kotlin 2.0. I will attempt to add some sort of coverage to prevent future regressions in a separate changeset.

Partially addresses #515


gradle/libs.versions.toml
```diff
@@ -18,15 +18,15 @@ nexus = "2.0.0"
 junit = "6.0.3"
 koverGradlePlugin = "0.9.8"
 downloadPlugin = "5.7.0"
-androidxBenchmark = "1.5.0-alpha06"
```
bidetofevil, line 21: This changed between alphas? yikes....

bidetofevil (approved):
LGTM

breedx-splk (approved):

gradle/libs.versions.toml
breedx-splk: If you look below it was `alpha05`, kinda like rc1 rc2 whatever.

breedx-splk:
This is a bummer, but better to actually be compatible....

I bet renovate will come back around, in which case we will need to be careful about merging until those regression tests are in place.

PR #521, opened 2026-05-20
Add test that enforces minimum supported versions

### LINKED ISSUE #515: Minimum supported Kotlin version is higher than stated in documentation

fractalwrench:
## Goal

Adds a test fixture that is built using [Gradle TestKit](https://docs.gradle.org/current/userguide/test_kit.html) that compiles against our min supported versions. The test runs on CI and should therefore break compilation if a regression is introduced via transitive dependencies. I've tested this out by bumping the coroutines dependency to the latest and confirmed that the build failed.

Closes #515

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -28,13 +28,18 @@ jobs:
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
           distribution: 'adopt'
-          java-version: 21
+          java-version: |
+            17
+            21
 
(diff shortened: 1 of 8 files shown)


bidetofevil (approved):
LGTM

PR #525, opened 2026-05-21
Add a test that cross checks propagation

### LINKED ISSUE #463: Add tests to confirm context propagation is same between implementation/compat modules
Testing that the implementation and compat modules produce identical results for context propagation. This can likely be achieved by using golden files (that assert JSON output from export are the same for both).

Originally raised in #460

fractalwrench:
## Goal

Adds a test that cross checks propagation between the implementation and compat modules to assert that they work the same. Closes #463.


bidetofevil (approved):
LGTM

breedx-splk (approved):
Nice.

Issue #379, opened 2026-04-07
Create Counter API


fractalwrench:
Create a Counter API as described in the spec: https://opentelemetry.io/docs/specs/otel/metrics/api/

DavidGrath:
Please I'd like to work on this issue. I have some inquiries about implementation details:
*  Would it be okay to use the unsigned number variants that Kotlin supports to implicitly support the non-negative requirement?
* I've drafted out the attributes using `Map<String, AnyValue>` for now in place of a full `Attributes` implementation. Is that alright?

fractalwrench:
Thanks for offering to help @DavidGrath, I've assigned you the issue!

Unsigned number variants sounds great.

The approach we've taken to accepting attribute parameters so far can be seen in [Logger#emit](https://github.com/open-telemetry/opentelemetry-kotlin/blob/8ad3cd882865c0061d1c398689346575bdefce9c/api/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/Logger.kt#L58).

PR #528, opened 2026-05-22
Started work on Counter API implementation. References Issue #379

DavidGrath:
## Goal

Addresses #379

## Testing

Added new test `FloatCounterAdapterTest` as a starter. Ran `apiDump`


api/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/DoubleCounter.kt
```diff
@@ -0,0 +1,20 @@
+package io.opentelemetry.kotlin.metrics
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.attributes.AttributesMutator
+
+
+@OptIn(ExperimentalApi::class)
+public interface DoubleCounter {
```
fractalwrench: Question to be discussed with the SIG: do we want to follow the approach of `<type-name>Counter` or using generics with `Counter<T>`. The spec gives wiggle room [on this](https://opentelemetry.io/docs/specs/otel/metrics/api/#counter-creation).

My first impression is that generics would be helpful in reducing the API surface but there may be hidden reasons why that's a bad idea.

api/api/jvm/api.api
```diff
@@ -219,7 +217,66 @@ public final class io/opentelemetry/kotlin/logging/SeverityNumber : java/lang/En
 	public static fun values ()[Lio/opentelemetry/kotlin/logging/SeverityNumber;
 }
 
+public abstract interface class io/opentelemetry/kotlin/metrics/DoubleCounter {
```
fractalwrench: Another question for the SIG: the spec states that most Instruments share a common set of properties. Does it make sense to define an `Instrument` interface that contains these and have `Counter` extend that, or are there reasons why opentelemetry-java hasn't followed that approach? https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument

api/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/DoubleCounter.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    public fun getName(): String
+
+    public fun getUnit(): String?
+    public fun getDescription(): String?
```
fractalwrench: nit: IMO `val name: String` would be preferable for any property whose value is guaranteed not to change

compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/metrics/DoubleCounterAdapter.kt
```diff
@@ -0,0 +1,52 @@
+package io.opentelemetry.kotlin.metrics
+
+import io.opentelemetry.kotlin.aliases.OtelJavaDoubleCounter
+import io.opentelemetry.kotlin.aliases.OtelJavaMeter
+import io.opentelemetry.kotlin.attributes.AttributesMutator
+import io.opentelemetry.kotlin.attributes.CompatAttributesModel
+
+internal class DoubleCounterAdapter(
+    private val name: String, private val unit: String?,
```
fractalwrench:
```suggestion
    private val name: String,
    private val unit: String?,
```

api/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/Meter.kt
```diff
@@ -15,4 +15,40 @@ import io.opentelemetry.kotlin.ThreadSafe
...
-public interface Meter
+public interface Meter {
+    /**
+     * @param name
+     * * Must not be empty
+     * * Maximum length of 255 characters
+     * * Must start with an alphabetic character
+     * * Consists of alphabets, numbers, and the following: '_', '.', '-', '/'
+     */
+    public fun createIntegerCounter(name: String, description: String? = null, unit: String? = null): IntegerCounter
```
fractalwrench: This signature seems to be missing `kind` as declared in the spec: https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument

Is that intentional by relying on `IntegerCounter` to suggest the type, or do we need to provide the `kind` in some other way?

fractalwrench (commented):
Many thanks for starting this off. I had some questions that I don't know the answer to myself and think would be good to discuss at the next SIG meeting, as it will influence what the correct API looks like for metrics as a whole.

api/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/Meter.kt
DavidGrath: Unintentional, my bad. I just missed it. I'll add it in the next commit

api/api/jvm/api.api
DavidGrath: I'm leaving this here so I don't forget, but I did some digging, and apparently they did go that route before diverging in [2021](https://github.com/open-telemetry/opentelemetry-java/pull/3423)

https://github.com/open-telemetry/opentelemetry-java/pull/3423/changes#diff-307abb73380afd863d0a12d2d2506cafe24a3577dd34071a8d3778ceb5173bce

It's a relatively fat PR, and the meeting notes only go back to 2024, so I'll need to look some more to try and find context behind the decision

DavidGrath: [Remove marker interfaces](https://github.com/open-telemetry/opentelemetry-java/pull/3423/changes/5d86d4b43ad28ba357802a68fd11384563e37acb)

As for now, my best guess is that it was to allow for the builder pattern to work well. 
I thought maybe the spec was significantly different back then, but nope, the same 4 properties were present in the latest version, [1.5.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.5.0/specification/metrics/api.md), before that PR was opened.  
I'll stop here and leave it up for discussion in the SIG meeting

DavidGrath: Relevant tag link for reference: [OTel Java 1.4.1](https://github.com/open-telemetry/opentelemetry-java/blob/v1.4.1/api/metrics/src/main/java/io/opentelemetry/api/metrics/DoubleCounter.java)

DavidGrath:
Ah, I see, I broke `MeterImpl`. This will take a while. I'm contemplating using the Noop module, but I think that's the wrong way to use it

DavidGrath:
It seems I can't mark it for review on the mobile app. I'll change the status when I can

DavidGrath:
So, in general, I started off trying to implement all 4 numeric types, but it started to get excessive especially when I needed to factor in `MeterImpl`, so I trimmed it down where I could using just `Long` while also trying to show what a complete implementation might look like in places like `Noop`

fractalwrench:
Closing this as I don't think it's actively being worked on - feel free to reopen if needed.

PR #551, opened 2026-05-22
feat(probability-sampling): variable precision threshold encoding

wpessers:
## Goal

This is just an initial refactor I want to propose before adding on the actual variable precision encoding / decoding logic for the threshold. Open to suggestions, also open to throwing away the entire refactor if maintainers feel like it's worse than before. Let me know!


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/Randomness.kt
```diff
@@ -0,0 +1,32 @@
+package io.opentelemetry.kotlin.tracing.sampling
+
+import io.opentelemetry.kotlin.factory.isValidLowercaseHex
+
+internal data class Randomness(internal val value: Long) {
```
fractalwrench: Have you considered whether a [value class](https://kotlinlang.org/docs/inline-classes.html) is possible? That would eliminate your concerns about extra allocations.

fractalwrench (commented):
I'm pretty happy with this approach and agree that it enhances the code's readability. I do think it'd be worth exploring a value class as that would avoid the extra allocations, but otherwise this looks good.

I'm out for the next few weeks so to move this forward you'll want to get further reviews from @bidetofevil or @open-telemetry/kotlin-approvers

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/Randomness.kt
wpessers: Good point, looks like the right fit as both `Randomness` and `Threshold` are basically each just wrapping one `Long`. Believe it would offer some gain, there's some cases here where I think it doesn't though. 2 reasons for that:

1. Nullability, IIUC the linked [docs](https://kotlinlang.org/docs/inline-classes.html#representation) "inline classes are boxed whenever they are used as another type". It looks like cases where `Randomness` or `Threshold` is nullable (`Threshold?` or `Randomness?`) would mean the inline classes would always be boxed and thus an allocation happens. There's 2 call sites in `ProbabilitySampler` where that would be the case:
    - `val explicitRandomness = otelTraceState.rv`, `OtelTraceState.rv` calls `Randomness.decode()`. The latter has a nullable return type `Randomness?` so I suppose that would always be boxed and thus cost an allocation regardless of value class or not.
    - `val incomingTh = otelTraceState.th`, `OtelTraceState.rv` calls `Threshold.decode()` same story as above.
2. Call sites where `Threshold` is treated as `Comparable<Threshold>` would lead to boxing, but there are none in the current code and I don't see why there would ever be. But to be 100% ceratin, I think we could just remove the interface implementation entirely on the `Threshold` class and change the `compareTo` function to be `operator fun compareTo...` to avoid the boxing risk here entirely.

wpessers: Use inline value class for both and removed the interface implementation for `Threshold`, so allocations should now be down to the minimum.

wpessers:
Excuse the force pushing, formatting fix only changing the most recent commit. No history rewrite so review history should not be broken.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/Randomness.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    companion object {
+        fun decode(encoded: String): Randomness? {
+            return encoded.takeIf { it.length == 14 && it.isValidLowercaseHex() }
+                ?.toLong(16)
+                ?.let(::Randomness)
+        }
+
+        fun fromTraceId(traceId: String): Randomness {
+            return Randomness(
```
breedx-splk, line 21: I think we'd want to guard against the array size, right? I've seen 8-byte trace IDs in the wild. 🤦🏻

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/Randomness.kt
```diff
@@ -0,0 +1,35 @@
...
+                (byteFromBase16(traceId[18], traceId[19]) shl 48) or
+                    (byteFromBase16(traceId[20], traceId[21]) shl 40) or
+                    (byteFromBase16(traceId[22], traceId[23]) shl 32) or
+                    (byteFromBase16(traceId[24], traceId[25]) shl 24) or
+                    (byteFromBase16(traceId[26], traceId[27]) shl 16) or
+                    (byteFromBase16(traceId[28], traceId[29]) shl 8) or
+                    byteFromBase16(traceId[30], traceId[31])
```
breedx-splk, line 28: Is there a reason to use this over `String.toLong(16)`? ([ref](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/to-long.html))?

breedx-splk (approved):
This is a good change, and we can always continue improving it, so I'm inclined to merge it, but did have a couple small questions. Thanks for the help!

wpessers:
Taking a look ASAP. I have been busier than expected and this PR was supposed to also include the actual variable precision encoding feature. But I just haven't gotten to that yet. If desired we could merge after I address the remaining remarks and I'll tale up the variable precision change in a follow-up PR...

breedx-splk:
> Taking a look ASAP. I have been busier than expected and this PR was supposed to also include the actual variable precision encoding feature. But I just haven't gotten to that yet. 

No worries, we're all super busy and there was a lag even responding to this. :)

> If desired we could merge after I address the remaining remarks and I'll tale up the variable precision change in a follow-up PR...

That works for me. I like smaller PRs in general. :) Thanks.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/Randomness.kt
wpessers: It is **possible** to indeed do something like `traceId.substring(18, 32).toLong(16)` instead of this, it would indeed give the exact same result. However, I've tried my best in the original implementation to optimize performance.

Be aware that I haven't introduced this logic in the current PR, this refactor just moves it. It existed previously in https://github.com/wpessers/opentelemetry-kotlin/blob/73a70428bb7dc3008984b24cce918aed831910bf/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/ProbabilitySampler.kt#L80-L87

Taking some inspiration from the java sdk: https://github.com/open-telemetry/opentelemetry-java/blob/a5cd87f127cd8d4e5845bc1e29360f815a6f01c8/api/all/src/main/java/io/opentelemetry/api/internal/OtelEncodingUtils.java#L56-L72
This style avoids the extra substring allocation. On its own I guess that would be negligible, but I just did this as part of a bigger effort to optimize the probability sampling code.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/Randomness.kt
wpessers: @breedx-splk I can add a simple check here, just to make sure the actual way "external" trace ids enter the sdk is always through either https://github.com/open-telemetry/opentelemetry-kotlin/blob/a1d0ab221a61795157ce4aa125d04523087b3b64/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceParent.kt#L75 or https://github.com/open-telemetry/opentelemetry-kotlin/blob/a1d0ab221a61795157ce4aa125d04523087b3b64/api/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/SpanContextFactory.kt#L28
Correct?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/Randomness.kt
breedx-splk: Thanks for the detailed clarification. I wasn't expecting the optimization this early...but I think it's just fine.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/Randomness.kt
breedx-splk: I think so, but let's not forget how software will morph and evolve over time. :)

breedx-splk: I think this should be the last thing and then we can get it merged.

breedx-splk:
@wpessers can we get that final guard in there and wrap this up? Thanks again!

fractalwrench:
Closing as I believe this has been overtaken by other work on main.

PR #568, opened 2026-05-26
Update RELEASING.md instructions

breedx-splk:
link directly to the clickable workflows, rather than the source. This makes it much simpler for the person doing the release.


bidetofevil (approved):
LGTM

PR #569, opened 2026-05-26
[do not merge] Split incubating semconv

breedx-splk:
This splits semconv into two packages: incubating and non-incubating. This is primarily an exercise so that we can discuss this approach, which more closely aligns with the existing java patterns (which we can certainly diverge from, pros/cons).


breedx-splk:
I'm content with the annotations for now, so I'm closing this. We'll have it as an artifact if we choose to revisit this idea later. ✌🏻

PR #573, opened 2026-06-06
fix: Use TraceService response model for OTLP trace response

### LINKED ISSUE #557: Use TraceService response model for OTLP trace response
Currently the LogService response model is used mistakenly. We should use the TraceService response model instead: https://github.com/open-telemetry/opentelemetry-kotlin/blob/53703f66033cec9922341b5fbb9a2434278df511/exporters-protobuf/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/ExportTracesResponseDeserializer.kt#L5

narumichi0710:
Use the correct ExportTraceServiceResponse model in ExportTracesResponseDeserializer instead of　ExportLogsServiceResponse.

Fixes #557


breedx-splk (approved):
Thanks!

PR #574, opened 2026-06-06
fix: Retry OTLP HTTP requests on retryable status codes

### LINKED ISSUE #559: Retry failed OTLP HTTP requests
The OTLP exporters should handle throttling with exponential backoff as per the spec: https://opentelemetry.io/docs/specs/otlp/#otlphttp-throttling

narumichi0710:
Retry OTLP HTTP requests on retryable status codes (429, 502, 503, 504) and honour the Retry-After header.

Fixes #559

### DIFF
--- a/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpClient.kt
+++ b/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpClient.kt
@@ -14,6 +14,7 @@ import io.ktor.http.contentType
 import io.ktor.utils.io.readRemaining
 import io.opentelemetry.kotlin.export.OtlpClient.Companion.MAX_ERROR_BODY_BYTES
 import io.opentelemetry.kotlin.export.OtlpResponse.ClientError
+import io.opentelemetry.kotlin.export.OtlpResponse.RetryableError
 import io.opentelemetry.kotlin.export.OtlpResponse.ServerError
 import io.opentelemetry.kotlin.export.OtlpResponse.Success
 import io.opentelemetry.kotlin.export.OtlpResponse.Unknown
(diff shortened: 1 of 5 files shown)


breedx-splk (approved):
This is a step in the right direction, but I don't think it fully closes the issue. The issue points to the spec which says that the client needs to perform its own exponential backoff when the Retry-After heading is missing. I'm not seeing where that happens.

breedx-splk:
@open-telemetry/kotlin-approvers what do the rest of you think? Ok to proceed?

narumichi0710:
@breedx-splk sorry for the late reply. The backoff is in TelemetryExporter.kt:

```kotlin
is OtlpResponse.RetryableError -> {
  delay(response.retryAfterMs ?: delayMs)
  delayMs = (delayMs * 2).coerceAtMost(maxAttemptIntervalMs)
}
```

When Retry-After is missing we delay by delayMs, then double it for the next attempt (capped at　maxAttemptIntervalMs). It's inline alongside the other retry cases. 
I can add a test that asserts the backoff progression to make this more explicit. Want me to go ahead, or did you have something else in mind?

breedx-splk:
> @breedx-splk sorry for the late reply. The backoff is in TelemetryExporter.kt:
> 
> ```kotlin
> is OtlpResponse.RetryableError -> {
>   delay(response.retryAfterMs ?: delayMs)
>   delayMs = (delayMs * 2).coerceAtMost(maxAttemptIntervalMs)
> }
> ```
> 
> When Retry-After is missing we delay by delayMs, then double it for the next attempt (capped at　maxAttemptIntervalMs). It's inline alongside the other retry cases. I can add a test that asserts the backoff progression to make this more explicit. Want me to go ahead, or did you have something else in mind?

Oh no that's completely fine. I think I just didn't see that or had overlooked it. Sorry!

PR #579, opened 2026-06-08
Add component matrix for api stability levels

breedx-splk:
As we discussed in SIG today, let's track individual component "stability" in a matrix here, until we can hit 1.0 or use suffixes or whatever.


jdvr (approved):

Msksgm (approved):
LGTM

PR #582, opened 2026-06-09
fix: TraceFlags.isRandom as true

### LINKED ISSUE #552: Set random flag on TraceFlags correctly
The random flag is set as false when creating TraceFlags: https://github.com/open-telemetry/opentelemetry-kotlin/blob/9f69ec568ede2ec4a75124f573ed6ab825a35144/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt#L110

However, we generate IDs that meet the specification's standard for randomness so the flag should be set: https://opentelemetry.io/docs/specs/otel/trace/sdk/#random-trace-flag

Msksgm:
## Goal

To conform to the random trace flag OpenTelemetry Specification.
This is because the traceIdBytes is generated from `kotlin.random.Random`.

Close https://github.com/open-telemetry/opentelemetry-kotlin/issues/552

## Testing

I updated some tests as the spanContext.traceFlags.hex expects "03".
I followed the W3C trace flags specification https://www.w3.org/TR/trace-context-2/#trace-flags .

- "00" → not Sampled & not Random
- "01" → Sampled & not Random
- "02" → not Sampled & Random
- "03" → Sampled & Random

### DIFF
--- a/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/tracing/ext/TraceFlagsExt.kt
+++ b/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/tracing/ext/TraceFlagsExt.kt
@@ -4,18 +4,11 @@ import io.opentelemetry.kotlin.aliases.OtelJavaTraceFlags
 import io.opentelemetry.kotlin.tracing.TraceFlags
 
 internal fun TraceFlags.toOtelJavaTraceFlags(): OtelJavaTraceFlags {
-    val sb = StringBuilder()
-    sb.append(
-        when {
-            isRandom -> "1"
-            else -> "0"
-        }
-    )
-    sb.append(
-        when {
-            isSampled -> "1"
-            else -> "0"
-        }
-    )
-    return OtelJavaTraceFlags.fromHex(sb.toString(), 0)
+    val traceFlagsHex = when {
+        isRandom && isSampled -> "03"
+        isRandom && !isSampled -> "02"
+        !isRandom && isSampled -> "01"
+        else -> "00"
+    }
+    return OtelJavaTraceFlags.fromHex(traceFlagsHex, 0)
 }
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/TraceFlagsFactoryImpl.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/TraceFlagsFactoryImpl.kt
@@ -6,7 +6,7 @@ import io.opentelemetry.kotlin.tracing.TraceFlagsImpl
 
 @ExperimentalApi
 internal class TraceFlagsFactoryImpl : TraceFlagsFactory {
-    override val default: TraceFlags by lazy { TraceFlagsImpl(isSampled = true, isRandom = false) }
+    override val default: TraceFlags by lazy { TraceFlagsImpl(isSampled = true, isRandom = true) }
 
     override fun fromHex(hex: String): TraceFlags {
         if (!hex.isValid()) {
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt
@@ -118,7 +118,7 @@ internal class TracerImpl(
             spanIdBytes = spanIdBytes,
             traceFlags = when {
                 sampled -> traceFlagsDefault
-                else -> TraceFlagsImpl(isSampled = false, isRandom = false)
+                else -> TraceFlagsImpl(isSampled = false, isRandom = true)
             },
             isValid = true,
             isRemote = false,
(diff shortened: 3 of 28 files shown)


Msksgm:
@open-telemetry/kotlin-maintainers 


While working on this PR, I noticed a difference between the test expectations in `compat` and `implementation`.Tests in `./implementation/src/jvmTest` expect `spanContext.traceFlags` to be `"03"`, while tests in `./compat/src/jvmTest` expect `"01"` or `"02"`.

After investigating, I found that the difference comes from the ID generator being used.`OtelKotlinHarness.kt` uses the Java SDK's `IdGenerator`, which has a default implementation of `generatesRandomTraceIds()` that returns `false`. As a result, the random flag is not set and the trace flags become `"01"` or `"02"`.

https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/framework/OtelKotlinHarness.kt#L33

https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/java-typealiases/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/aliases/Aliases.kt#L137

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/IdGenerator.java#L42-L53

Meanwhile, `IntegrationTestHarness` uses the Kotlin SDK's `IdGeneratorImpl`, which always sets the random flag and therefore produces `"03"`.


https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/IdGeneratorImpl.kt

https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/sdk-api/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/IdGenerator.kt

Because of this, I'm not sure what the intended behavior should be.

1. Update `OtelKotlinHarness.kt` so that `generatesRandomTraceIds()` returns `true`, making the compat tests consistent with the implementation tests?
2. Update `IdGeneratorImpl.kt` so that it does not set the random flag?

Could someone clarify which approach is expected here?

Thanks!

bidetofevil:
Feels like if we are using something called `FakeIdGenerator` (which defers to the Java impl), we shouldn't be validating the correctness of the IDs generated. Which test are asserting this but somehow calls into a fake impl of `IdGenerator`?

Msksgm:
@bidetofevil 

Thanks for the review!

The `traceFlags` value is asserted in two places:

1. **Golden JSON comparison** — `compareGoldenFile` does a structural equality check on the serialized output, which includes `"traceFlags"`. So tests using `goldenFileName = "..."` (`SpanExportTest`, `OtelJavaSpanExportTest`, etc.) implicitly assert it. See [`SerializableSpanContext.kt#L11`](https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/integration-test/src/commonMain/kotlin/io/opentelemetry/kotlin/framework/serialization/conversion/SerializableSpanContext.kt#L11) (`traceFlags = traceFlags.hex`).
2. **In-memory assertions** — a few tests check `traceFlags` directly:
   - [`SpanExportTest.kt#L137-143`](https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/tracing/SpanExportTest.kt#L137-L143)
   - [`OtelJavaSpanExportTest.kt#L132-138`](https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/tracing/OtelJavaSpanExportTest.kt#L132-L138)

Would it be acceptable to add `override fun generatesRandomTraceIds(): Boolean = true` to `FakeIdGenerator` in
[`OtelKotlinHarness.kt`](https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/framework/OtelKotlinHarness.kt#L31-L55)?

Java SDK's own `RandomIdGenerator` already overrides this to `true` ([`RandomIdGenerator.java#L41-44`](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RandomIdGenerator.java#L41-L44)), so this would just align the test fake with the production behavior (the expected `traceFlags` becomes `"03"`, matching the Kotlin SDK tests).

fractalwrench (commented):
@Msksgm thanks for opening this. I think that including `public val supportsRandomTraceIds: Boolean = true` on `IdGenerator` would be helpful, and each implementation can set the value appropriately.

Additionally, within `calculateSpanContext` it is necessary to respect the parent's value for `isRandom` if a parent trace ID exists, as this could be different from the local generated value.

There are a couple of test cases to add around this, but asides from that this looks like it's nearly ready to merge.

Msksgm:
Thanks for taking care of this in #755. I'm sorry I wasn't able to follow up on this PR in time.

I appreciate the maintainers picking this up and getting the fix in.

PR #585, opened 2026-06-11
Event semantic generation

breedx-splk:
Well, it's a lot of weaver generated code, so naturally it's quite large.

So this adds a weaver template for events. Each event type is a class that encapsulates the event name and enforces required attributes via non-null constructor arguments, while keeping optional attributes as defaulted nullable arguments.

Usage generally looks something like this:

```kotlin
SessionStartEvent(
      sessionId = "session-123",
      sessionPreviousId = "session-122",
  ).emit(logger) {
      setStringAttribute("bogo.app.user", "jimmy666")
      setBooleanAttribute("bogo.app.milky", false)
  }
```

The main points being:
* emit takes a logger, the logger will already have a scope set.
* additional attributes may be set by implementing a lambda that receives an `AttributesMutator`.


semconv/src/commonMain/kotlin/io/opentelemetry/kotlin/semconv/events/OpenTelemetryEvent.kt
```diff
@@ -0,0 +1,13 @@
+package io.opentelemetry.kotlin.semconv.events
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.attributes.AttributesMutator
+import io.opentelemetry.kotlin.logging.Logger
+
+@OptIn(ExperimentalApi::class)
```
priettt:
```suggestion
@ExperimentalApi
```

Shouldn't this be marked as Experimental too?

priettt (approved):
Looks good 🙌  

Left a comment about @ExperimentalApi

semconv/src/commonMain/kotlin/io/opentelemetry/kotlin/semconv/events/OpenTelemetryEvent.kt
breedx-splk: Yes! Good catch, thanks, I'll fix this.

PR #588, opened 2026-06-12
chore(docs): Minor improvements for CONTRIBUTING.md

jdvr:
## Goal

Small improvements gathered after a first run of the project locally. No functional changes to the library. No functional code changed.

**`CONTRIBUTING.md` improvements:**
- Clarify that the full Xcode app is required (not just command-line tools), and add setup commands
- Link to the Kotlin Multiplatform compatibility guide for supported Xcode versions, and note the two options when your Xcode is newer than the supported range (install an older iOS platform runtime via Xcode → Settings → Platforms, or temporarily exclude iOS tests)
- List the specific Android SDK packages required (`platforms;android-34`, `build-tools;30.0.3`, `platform-tools`) rather than just linking to Android Studio
- Document that `publishToMavenLocal` must be run before `./gradlew build` when the `gradle-integration-test` module is included, and explain why (Kotlin/Native race condition)

**`PublishConfig.kt`:**
- `signAllPublications()` was unconditional, causing `publishToMavenLocal` to fail for contributors without a GPG key. Changed to only sign when `signingKey` or `signing.keyId` properties are present, which matches the pattern used by most OSS projects.

## Testing

- Ran `./gradlew publishToMavenLocal` successfully without a GPG key configured
- Ran `./gradlew build` with the integration test passing after aligning the snapshot version

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,11 +8,41 @@ would be accepted.
 
 1. Fork and clone the repository
 2. Install the following prerequisites:
-   1. JDK >=11 (OpenJdk 21 using https://sdkman.io/ is recommended)
-   2. Android SDK: https://developer.android.com/studio
-   3. Android Studio or IntelliJ IDEA are recommended
+   1. JDK >=11 (OpenJDK 21 via https://sdkman.io/ is recommended)
+   2. **Android SDK** — Android Studio bundles the SDK and is the easiest route: https://developer.android.com/studio.
+      If installing the SDK manually, ensure the following packages are present:
+      - `platforms;android-34`
+      - `build-tools;30.0.3`
+      - `platform-tools`
+   3. **Xcode** (macOS only) — the full Xcode app is required, not just the command-line tools.
+      Install from the Mac App Store, then run:
+      ```bash
+      sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
+      sudo xcodebuild -license accept
+      ```
+      The supported Xcode version depends on the Kotlin version used in the project — check the
+      [Kotlin Multiplatform compatibility guide](https://kotlinlang.org/docs/multiplatform/multiplatform-compatibility-guide.html).
+      If your Xcode version is newer than the supported range, iOS simulator tests will fail; either
+      install a supported iOS platform runtime via **Xcode → Settings → Platforms**, or exclude the
+      tests temporarily with `-x iosSimulatorArm64Test -x iosArm64Test`.
+   4. Android Studio or IntelliJ IDEA are recommended
 3. Run `./gradlew build` to confirm the project builds
 4. Open an issue or update these docs if there was a step missing from these instructions!
+
+### Running the full build including the integration test
+
+The `gradle-integration-test` module verifies that the library's published artifacts are consumable
+by the minimum supported Gradle version. It requires the snapshot artifacts to be published to
+Maven Local first. This is a separate step because running both in the same Gradle invocation causes
+a race condition with Kotlin/Native:
+
+```bash
+./gradlew publishToMavenLocal   # publishes snapshot artifacts to ~/.m2
+./gradlew build                  # runs all tests including the integration test
+```
+
+No GPG signing key is required for local development — signing is skipped automatically when no
+key is configured.
 
 ## Development guidelines
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,8 +41,8 @@ a race condition with Kotlin/Native:
 ./gradlew build                  # runs all tests including the integration test
 ```
 
-No GPG signing key is required for local development — signing is skipped automatically when no
-key is configured.
+No GPG signing key is required for local development — pass `-Psigning.skip=true` to skip signing,
+or add `signing.skip=true` to your local `gradle.properties`.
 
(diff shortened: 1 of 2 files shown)


jdvr:
❓ When running `publishToMavenLocal` with `version=0.5.0` in `gradle.properties`, artifacts are published as `0.5.0`, but `gradle-integration-test` always appends `-SNAPSHOT` (`val snapshotVersion = "${project.version}-SNAPSHOT"`) [ref](https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/gradle-integration-test/build.gradle.kts#L19), causing the integration test to fail to resolve the artifacts. Is the expectation that contributors always set `version=0.5.0-SNAPSHOT` locally, or should the integration test be more flexible about the version suffix?

CONTRIBUTING.md
```diff
@@ -8,12 +8,42 @@ would be accepted.
...
 2. Install the following prerequisites:
-   1. JDK >=11 (OpenJdk 21 using https://sdkman.io/ is recommended)
-   2. Android SDK: https://developer.android.com/studio
-   3. Android Studio or IntelliJ IDEA are recommended
+   1. JDK >=11 (OpenJDK 21 via https://sdkman.io/ is recommended)
+   2. **Android SDK** — Android Studio bundles the SDK and is the easiest route: https://developer.android.com/studio.
+      If installing the SDK manually, ensure the following packages are present:
+      - `platforms;android-34`
+      - `build-tools;30.0.3`
+      - `platform-tools`
```
jdvr, line 16: I am not sure if this is 100% true, I am not familiar with android sdk set up and I added sdks as I saw the log failures following AI instructions on how to configure it.

DavidGrath: Hello Juan, I'd like to help  out a bit

I tried setting up Android in WSL once - it's a weird story. Based on my notes to myself, if the emulator is also needed, then this should also be considered:

Check for existence
```bash
sdkmanager --list_installed
```

Look for
```
"system-images;android-21;google_apis;x86_64"
"emulator"
```
Install
```bash
sdkmanager "system-images;android-21;google_apis;x86_64" "emulator" "platform-tools"
```
## Emulator
Check for existence
```bash
avdmanager list avd
```

Create
```bash
avdmanager create avd --name TestDefault21 -k "system-images;android-21;google_apis;x86_64" -d pixel_xl
```
Then this is a quirky one, too:
```bash
apt-get install libpulse0
```

DavidGrath: I kept the version at 21 because that's what I used at the time, but it can be changed if needed
My idea was setting up a CI server on bare metal, so maybe it's not 1:1 applicable

buildSrc/src/main/kotlin/io/opentelemetry/kotlin/PublishConfig.kt
```diff
@@ -9,7 +9,9 @@ fun Project.configurePublishing() {
 
         mavenPublishing.apply {
             publishToMavenCentral(automaticRelease = true)
-            signAllPublications()
+            if (project.hasProperty("signingKey") || project.hasProperty("signing.keyId")) {
```
priettt: I think this guard will skip signing during releases, which will fail Maven Central validation (release artifacts must be PGP-signed).

The release workflow provides the key via `ORG_GRADLE_PROJECT_signingInMemoryKey` (and ...KeyId/...KeyPassword) in [release.yml](https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/.github/workflows/release.yml#L7-L9). 

At release time, this will create a property called `signingInMemoryKey`, not `signingKey` or `signing.keyId`. [Gradle docs](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables)

priettt (commented):
Left a comment about the signing guard, otherwise it looks good 👍

buildSrc/src/main/kotlin/io/opentelemetry/kotlin/PublishConfig.kt
jdvr: Would it be better to have a disable property? like `signing.skip`?

priettt: That'd work too, we'd have that enabled as a default and only set it in release.yml. 

Looking for the already existing signingInMemoryKey is easier, but a property owned by us would cover us from changes in that property's name. I think both approaches are fine 👍

buildSrc/src/main/kotlin/io/opentelemetry/kotlin/PublishConfig.kt
```diff
@@ -9,7 +9,9 @@ fun Project.configurePublishing() {
 
         mavenPublishing.apply {
             publishToMavenCentral(automaticRelease = true)
-            signAllPublications()
+            if (!project.hasProperty("signing.skip")) {
```
jdvr, line 12: @priettt I have changed this!

CONTRIBUTING.md
jdvr: I tried to keep it minimal. I think further investment on WSL setup could be added as a separate section, but I've never worked with WSL and don't have the context to do it justice.

breedx-splk (approved):
👍🏻

PR #589, opened 2026-06-14
fix: span must not be exported if isSampled is false

### LINKED ISSUE #550: Respect RECORD_ONLY result decision for Sampler
The `Sampler` correctly determines whether a `Span` should be recorded: https://github.com/open-telemetry/opentelemetry-kotlin/blob/9f69ec568ede2ec4a75124f573ed6ab825a35144/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt#L76

However, if the decision is `RECORD_ONLY` the `Span` is still passed to the span processor. We should alter this and add test coverage to ensure this is not the case.

Msksgm:
## Goal

Close https://github.com/open-telemetry/opentelemetry-kotlin/issues/550

Previously, when the `Sampler` returned `RECORD_ONLY` (i.e. `SpanContext.traceFlags.isSampled == false`), the span was still forwarded to `SpanExporter`s. This violates the OpenTelemetry [Built-in span processors](https://opentelemetry.io/docs/specs/otel/trace/sdk/#built-in-span-processors) specification.

This change makes every `SpanProcessor.onEnd` short-circuit when the span is not sampled, so unsampled spans are no longer passed to the exporter.

### Changes

- Added an `isSampled` early-return at the top of `onEnd` in the following `SpanProcessor` implementations:
  - `SimpleSpanProcessor`
  - `BatchSpanProcessorImpl`
  - `PersistingSpanProcessor`
  - `InMemorySpanProcessor` (both in the `integration-test` and `testing` modules)
- Changed the default of `FakeTraceFlags.isSampled` to `true`. Existing tests assumed `isSampled == true`, but the previous non-compliant implementation also exported unsampled spans, so the tests happened to pass with `isSampled = false` as well. Once the spec-compliant behavior is in place, the test fake needs this default to keep the existing test intent.
- Adjusted the expected hex values in `TraceFlagsExtTest` and the `traceFlags` field in the JSON test resources from `"00"` to `"01"` to match the new default.

## Testing

- Added unit tests to `SimpleSpanProcessorTest` and `BatchSpanProcessorImplTest` that pass a span with `isSampled == false` to `onEnd` and assert that `FakeSpanExporter.exports` remains empty.
- Verified locally with `./gradlew build` (existing tests, new tests, and Detekt all pass).

### DIFF
--- a/api-ext/src/commonTest/kotlin/io/opentelemetry/kotlin/tracing/model/TraceFlagsExtTest.kt
+++ b/api-ext/src/commonTest/kotlin/io/opentelemetry/kotlin/tracing/model/TraceFlagsExtTest.kt
@@ -8,9 +8,9 @@ internal class TraceFlagsExtTest {
 
     @Test
     fun testEmptyString() {
-        assertEquals("00", FakeTraceFlags().hex)
-        assertEquals("01", FakeTraceFlags(isSampled = true).hex)
-        assertEquals("02", FakeTraceFlags(isRandom = true).hex)
-        assertEquals("03", FakeTraceFlags(isSampled = true, isRandom = true).hex)
+        assertEquals("00", FakeTraceFlags(isSampled = false).hex)
+        assertEquals("01", FakeTraceFlags().hex)
+        assertEquals("02", FakeTraceFlags(isSampled = false, isRandom = true).hex)
+        assertEquals("03", FakeTraceFlags(isRandom = true).hex)
     }
 }
--- a/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/BatchSpanProcessorImpl.kt
+++ b/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/BatchSpanProcessorImpl.kt
@@ -33,7 +33,12 @@ internal class BatchSpanProcessorImpl(
             exportAction = exporter::export
         )
 
-    override fun onEnd(span: ReadableSpan) = shutdownState.execute { processor.processTelemetry(span) }
+    override fun onEnd(span: ReadableSpan) {
+        if (!span.spanContext.traceFlags.isSampled) {
+            return
+        }
+        shutdownState.execute { processor.processTelemetry(span) }
+    }
 
     override fun isStartRequired(): Boolean = true
     override fun isEndRequired(): Boolean = true
--- a/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/SimpleSpanProcessor.kt
+++ b/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/SimpleSpanProcessor.kt
@@ -34,6 +34,9 @@ internal class SimpleSpanProcessor(
     }
 
     override fun onEnd(span: ReadableSpan) {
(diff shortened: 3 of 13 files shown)


exporters-persistence/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/PersistingSpanProcessor.kt
```diff
@@ -115,15 +115,17 @@ internal class PersistingSpanProcessor(
...
-        } catch (e: Throwable) {
-            sdkErrorHandler.onUserCodeError(
-                e,
-                "SpanProcessor.onEnd failed",
-                SdkErrorSeverity.WARNING
-            )
+    override fun onEnd(span: ReadableSpan) {
+        shutdownState.execute {
+            try {
+                composite.onEnd(span)
```
priettt: shouldn't we add the `isSampled` guard here too?

priettt: and also add a test for it

priettt (commented):
Looks good, though I think there's a missing guard

exporters-persistence/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/PersistingSpanProcessor.kt
Msksgm: I thought each processor already guards against exporting unsampled spans. Since the composite processor delegates to multiple processors, wouldn't this become a duplicate check?

Do we still need the `isSampled` guard here?

https://github.com/open-telemetry/opentelemetry-kotlin/pull/589/changes/BASE..48a091f6ca810c37cd05616f036334f5038a077d#diff-7a55785a946952834dddba245c8324ae9167df0e0604f5fe18676ff1a35073caL79

Msksgm: This diff (`=` → `return`) was my mistake.
If you agree that we don't need to add the `isSampled` guard here and that we don't need additional tests for it, I'll revert this change and restore the original implementation.

priettt: Sorry for the delay, you're right, the isSampled guard is not needed!

priettt (approved):

exporters-persistence/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/PersistingSpanProcessor.kt
Msksgm: Thank you!

Msksgm: I reverted the change.

breedx-splk (approved):
The implementation looks good. I wish it was a little clearer that this only applies to the built-in exporting span processors, and not to user-supplied or other random/helpful span processors. That is, `onEnd` is still always called when `isSampled=true`, but these exporting implementations bypass export.

PR #590, opened 2026-06-14
fix: ReadWriteLogRecord.SpanContext as variable

### LINKED ISSUE #553: Allow overwriting SpanContext on ReadWriteLogRecord
It should be possible to overwrite the `SpanContext` on `ReadWriteLogRecord` as logs may be generated from 3rd party libraries that need passing through a processor to add the appropriate context. Currently this is `val`.

Msksgm:
## Goal

Closes #553

Makes `spanContext` mutable on `ReadWriteLogRecord` so that `LogRecordProcessor` implementations can override it during `onEmit`, consistent with how other log record fields (`body`, `timestamp`, `severityNumber`, etc.) are already handled.

## Changes

- **`sdk-api`**: Declares `spanContext` as `var` in the `ReadWriteLogRecord` interface
- **`implementation`**: Updates `LogRecordModel` to store `spanContext` as a thread-safe `var` backed by a read-write lock, consistent with other mutable fields
- **`compat`**: Updates `ReadWriteLogRecordAdapter` to expose `spanContext` as `var`
- **`sdk-api` API dump**: Adds `getSpanContext` / `setSpanContext` to the public API surface

## Testing

- Extended `LogProcessorOnEmitTest.testOverridePropertiesInProcessor` to:
  - Assert the original `spanContext` (traceId, spanId, traceFlags, traceState) before the override
  - Override `spanContext` with `FakeSpanContext.VALID` in the processor
  - Verify the exported record reflects the new span context via golden file (`log_emit_override.json`)
- `./gradlew build` passes

## Known limitations

The `spanContext` setter in `ReadWriteLogRecordAdapter` (compat module) is a no-op, consistent with all other property setters in the same class. This is a limitation of the Java SDK interop layer — `OtelJavaReadWriteLogRecord` does not expose setters for these fields. Mutations in `LogRecordProcessor.onEmit` will have no effect when using the compat backend.

### DIFF
--- a/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/logging/export/ReadWriteLogRecordAdapter.kt
+++ b/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/logging/export/ReadWriteLogRecordAdapter.kt
@@ -44,6 +44,11 @@ internal class ReadWriteLogRecordAdapter(
         set(value) {
         }
 
+    override var spanContext: SpanContext
+        get() = SpanContextAdapter(impl.spanContext)
+        set(value) {
+        }
+
     override var eventName: String?
         get() = impl.eventName
         set(value) {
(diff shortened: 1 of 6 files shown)


priettt (approved):
LGTM 🙌

breedx-splk (approved):
Thanks for the contribution!

PR #595, opened 2026-06-15
fix: make setter/getter optional arguments on TextMapPropagator

### LINKED ISSUE #542: Make setter/getter optional arguments on TextMapPropagator
The Propagator API defines these as optional but currently these are mandatory in the interface. We should default them to null instead: https://opentelemetry.io/docs/specs/otel/context/api-propagators/#overview

Msksgm:
## Goal

The [Propagator API spec](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#overview) defines the carrier argument on `TextMapSetter` and `TextMapGetter` as optional (i.e. nullable). However, the current interface signatures make the carrier non-nullable, which is not compliant with the spec.

This PR updates `TextMapSetter.set` and `TextMapGetter.get` so that the carrier parameter is nullable (`C?`), aligning the API with the OpenTelemetry specification.

Close https://github.com/open-telemetry/opentelemetry-kotlin/issues/542

## Testing

All existing propagation tests continue to pass after the change. Implementations that previously assumed a non-null carrier (e.g. `MapTextMapGetter`, `MapTextMapSetter` in tests and compat adapters) have been updated to use safe-call operators (`?.`) on the nullable carrier, ensuring correct null-handling behaviour at call sites.

### DIFF
--- a/api/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TextMapGetter.kt
+++ b/api/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TextMapGetter.kt
@@ -21,7 +21,7 @@ public interface TextMapGetter<C> {
      * Returns the first value for [key] in [carrier], or null if absent.
      * Key lookup must be case-insensitive for HTTP carriers.
      */
-    public fun get(carrier: C, key: String): String?
+    public fun get(carrier: C?, key: String): String?
 
     /**
      * Returns all values associated with [key] in [carrier], in the order they appear,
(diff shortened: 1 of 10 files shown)


jdvr (approved):
This impacts #605 

LGTM 😄

priettt (approved):
Looks good 👍 

A small point: should we also modify getAll to make it follow the same structure?

carlosalberto:
IMHO not really in the Spec _but_ this is what Java does (in order to be more resilient to errors), so good to go.

Also the PR may give the wrong idea, as it's the `carrier` argument for both getter/setters that is optional.

PR #605, opened 2026-06-22
feat(Context API): Implement B3 Propagator

### LINKED ISSUE #459: Implement a `B3` propagator
Implement a `B3` propagator: https://opentelemetry.io/docs/specs/otel/context/api-propagators/#b3-requirements

jdvr:
## Goal

fixes #459 

### Add a B3Propagator implementing the [B3 propagation spec](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#b3-requirements)

- Always extracts both single (b3) and multi-header (X-B3-*) formats, with single taking precedence
- Inject format is configurable via `B3Format` (single by default)
- Preserves the B3 debug flag across hops via a ContextKey
- Accepts 64-bit (16-char) trace IDs by zero-padding to 128-bit
- Exposed through `PropagatorConfigDsl.b3` (including compat)

## Refactor

Replaced the per-type `DeferredW3CTraceContextPropagator` and `DeferredB3Propagator` classes with a single generic `ForwardingPropagator(delegate: () -> TextMapPropagator)` backed by `@Volatile` vars in `PropagatorConfigImpl`. I feel it is more idiomatic Kotlin.
Lambdas and `@Volatile` vars express the intent directly without boilerplate wrapper classes. I tried with `lazy {}` but ruled out since it initializes only once; re-reading the var on every call preserves the safe noop fallback if accessed before SDK init.

## Testing

- `B3PropagatorTest` covering:
  - inject/extract for both formats
  - debug flag round-trips
  - invalid input handling
  - 64-bit traceId padding
  - single-over-multi precedence
- `CorePropagatorApiTest` extended with b3() DSL tests
- `CompatB3PropagatorApiTest` verifying single/multi round-trips through the compat path

**Full build passing including iOS native compilation** (`./gradlew build`)


compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/propagation/CompatPropagatorConfigImpl.kt
```diff
@@ -28,6 +30,15 @@ internal class CompatPropagatorConfigImpl : PropagatorConfigDsl {
         return configured
     }
 
+    override fun b3(format: B3Format): TextMapPropagator {
```
jdvr, line 33: is this the right approach? I was hesitant between this or using the new implementation but that would break the separation of `implementation` and `compat` modules

compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/propagation/CompatPropagatorApiTest.kt
```diff
@@ -164,18 +164,3 @@ private class ContextWritingPropagator(
     override fun <T> extract(context: Context, carrier: T, getter: TextMapGetter<T>): Context =
         context.set(key, value)
 }
-
-@OptIn(ExperimentalApi::class)
```
jdvr, line 168: Moved to `MapCarrierFixtures` ([ref](https://github.com/open-telemetry/opentelemetry-kotlin/pull/605/changes#diff-9800933869d9ff36060e510927700d866e319fb9544cf210d116c9fb00bd8c9f))

compat/build.gradle.kts
```diff
@@ -21,6 +21,7 @@ kotlin {
                 implementation(libs.opentelemetry.api)
                 implementation(libs.opentelemetry.sdk)
                 implementation(libs.opentelemetry.sdk.extension.incubator)
+                implementation(libs.opentelemetry.extension.trace.propagators)
```
jdvr, line 24: This brings the `B3Propagator`

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/init/PropagatorConfigImpl.kt
```diff
@@ -20,7 +20,12 @@ import kotlin.concurrent.Volatile
 internal class PropagatorConfigImpl : PropagatorConfigDsl {
 
     private var configured: TextMapPropagator = NoopOpenTelemetry.propagator
-    private val w3cTraceContext: DeferredW3CTraceContextPropagator = DeferredW3CTraceContextPropagator()
```
jdvr, line 23: Instead of creating another `DeferredB3...`, I have unify the  approach in this config class and create an agnostic `ForwardingPropagator`.

jdvr:
> ❌ Patch coverage is `92.42424%` with `10 lines` in your changes missing coverage. Please review

I'll review this

compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/propagation/CompatB3PropagatorApiTest.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertSame
+import kotlin.test.assertTrue
+
+@OptIn(ExperimentalApi::class)
+internal class CompatB3PropagatorApiTest {
+
+    private val dsl = CompatPropagatorConfigImpl()
+    private val contextFactory = CompatContextFactory()
```
breedx-splk: nonblocking nit: I'd rather duplicate these in each test to get some additional guarantee of test separation, to have confidence that no state is being shared/borrowed between tests. Not a big deal, just a thought for future recommendations/cleanups.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/init/PropagatorConfigImpl.kt
breedx-splk: yeah I like the way this reads too. 👍🏻

compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/propagation/CompatPropagatorConfigImpl.kt
breedx-splk: This is a good question. I'm not sure. Maybe @bidetofevil or @fractalwrench have ideas.

If we consider the w3c propagator as our baseline reference, we should use the new implementation. That's what we do for w3c anyway, and I think it's safe, even if the implementations _could_ differ....

implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/factory/HexTest.kt
```diff
@@ -168,4 +169,16 @@ internal class HexTest {
     fun testMixedValidAndInvalidIsNotValidLowercaseHex() {
         assertFalse("0123456789abcdefg".isValidLowercaseHex())
     }
+
+    @Test
+    fun testIsAllZerosHex() {
+        assertEquals(true, "0000000000000000".isAllZerosHex(), "16-char all-zero span ID")
+        assertEquals(true, "00000000000000000000000000000000".isAllZerosHex(), "32-char all-zero trace ID")
```
breedx-splk: we should be using `assertTrue()` and `assertFalse()` for these, rather than `assertEquals()`

implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/factory/HexTest.kt
```diff
@@ -168,4 +169,16 @@ internal class HexTest {
...
+        assertEquals(false, "b7ad6b7169203331".isAllZerosHex(), "non-zero span ID")
+        assertEquals(false, "0af7651916cd43dd8448eb211c80319c".isAllZerosHex(), "non-zero trace ID")
+        assertEquals(false, "ffffffffffffffff".isAllZerosHex(), "max value span ID")
+        assertEquals(false, "ffffffffffffffffffffffffffffffff".isAllZerosHex(), "max value trace ID")
+        assertEquals(false, "00000000000000010000000000000000".isAllZerosHex(), "non-zero high half only")
+        assertEquals(false, "00000000000000000000000000000001".isAllZerosHex(), "non-zero low half only")
+    }
```
breedx-splk, line 183: add empty string case?

implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/propagation/B3PropagatorTest.kt
```diff
@@ -0,0 +1,329 @@
...
+        return mutableMapOf<String, String>().also { singlePropagator.inject(ctx, it, MapTextMapSetter) }
+    }
+
+    private fun injectMulti(sampled: Boolean): MutableMap<String, String> {
+        val flags = if (sampled) { traceFlagsFactory.fromHex("01") } else { traceFlagsFactory.fromHex("00") }
+        val spanContext = spanContextFactory.create(traceId, spanId, flags, traceStateFactory.default, false)
+        val ctx = contextFactory.root().storeSpan(spanFactory.fromSpanContext(spanContext))
+        return mutableMapOf<String, String>().also { multiPropagator.inject(ctx, it, MapTextMapSetter) }
+    }
+}
```
breedx-splk, line 329: Wow, thorough test and easy to read! Thanks!

breedx-splk (commented):
Thanks! This is mostly looking good, but I did leave a few small comments. Appreciate it!

compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/propagation/CompatPropagatorConfigImpl.kt
fractalwrench: I'm not sure what 'new implementation' means, could I have more context?

The function itself looks reasonable to me.

compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/propagation/CompatB3PropagatorApiTest.kt
fractalwrench: Agree. Either duplication or a setup function would be preferable

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/B3Propagator.kt
```diff
@@ -0,0 +1,186 @@
...
+    private fun <T> injectSingle(
+        spanContext: SpanContext,
+        debug: Boolean,
+        carrier: T,
+        setter: TextMapSetter<T>,
+    ) {
+        val flag = when {
+            debug -> 'd'
+            spanContext.traceFlags.isSampled -> '1'
+            else -> '0'
```
fractalwrench: nit: can these be extracted to constants?

fractalwrench (commented):
Thanks for opening this up @jdvr! I've added a couple of comments inline, once all comments are addressed & conflicts resolved, I don't see any reason why this couldn't be merged.

jdvr:
Thanks for all the comments. I will address them! 😄

compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/propagation/CompatPropagatorConfigImpl.kt
jdvr: I think this one is clear for me after reviewing the code. We don't want to have a dependency between imlementation and compat. So this is the right approach.

implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/factory/HexTest.kt
jdvr: I didn't add an empty-string case because I only treat 16 and 32-character strings as valid hex IDs. For other lengths, should this throw, default to false, or default to true?  I defaulted to false because it feels more natural with the question: "is this a valid all zeros hex id?". We could throw for invalid input but it feels like too much for this simple use case.
What do you think?

jdvr:
@fractalwrench @breedx-splk I have addressed all the comnents!

fractalwrench (approved):
LGTM

PR #607, opened 2026-06-27
feat: links supplied to sampler are an empty list

Msksgm:


Msksgm:
I mistook remote branch.
Then, I recreated it as https://github.com/open-telemetry/opentelemetry-kotlin/pull/608 .

PR #608, opened 2026-06-27
fix: links supplied to sampler are an empty list

### LINKED ISSUE #546: Links supplied to Sampler are an empty list
The links [supplied to `Sampler`](https://github.com/open-telemetry/opentelemetry-kotlin/blob/9f69ec568ede2ec4a75124f573ed6ab825a35144/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt#L70) are an empty list which means the links aren't visible to the sampler as [required in the spec](https://opentelemetry.io/docs/specs/otel/trace/api/#specifying-links).

Msksgm:
## Goal

Previously, `shouldSample` was called with empty attributes and links, so samplers had no way to use user-provided span attributes or links when making sampling decisions. This did not comply with the OpenTelemetry specification.

To fix this, `SpanCreationCollector` was introduced to collect user-provided attributes and links from the `action` lambda before sampling. The following changes were made:

- Introduce `SpanCreationCollector` and execute the `action` lambda before calling `shouldSample`
- Pass the collected attributes and links to `shouldSample` to inform the sampling decision
- Seed `SpanModel` with the collected links via a new `initialLinks` constructor parameter

Closes #546

## Testing

The following tests were added to `TracerSamplerTest` to verify that the sampler receives user-provided attributes and links:

- `testSamplerReceiverUserAttributes` — verifies that user attributes set in the `action` lambda are passed to `shouldSample`
- `testSamplerReceiversUserLinks` — verifies that user links added in the `action` lambda are passed to `shouldSample` in the correct order
- `testSamplerReceivesCappedLinks` — verifies that links are capped before being passed to `shouldSample`
- `testSamplerReceivesDedupedLinks` — verifies that duplicate links are deduplicated before being passed to `shouldSample`

Existing tests in `TracerSamplerTest` and `SpanLinkTest` were confirmed to remain green, ensuring no regression in sampler attribute application order or link handling behavior.

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/BuildSpanLink.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/BuildSpanLink.kt
@@ -21,8 +21,6 @@ internal fun buildSpanLink(
         attributeLimit = spanLimitConfig.attributeCountPerLinkLimit,
         attributeValueLengthLimit = spanLimitConfig.attributeValueLengthLimit
     )
-    if (attributes != null) {
-        attributes(container)
-    }
+    attributes?.invoke(container)
     return SpanLinkImpl(spanContext, container)
 }
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/BuildSpanLink.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/BuildSpanLink.kt
@@ -1,0 +1,22 @@
+package io.opentelemetry.kotlin.tracing.model
+
+import io.opentelemetry.kotlin.attributes.AttributesModel
+import io.opentelemetry.kotlin.attributes.AttributesMutator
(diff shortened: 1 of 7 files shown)


jdvr (approved):

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/BuildSpanLink.kt
```diff
@@ -0,0 +1,28 @@
...
+    attributes: (AttributesMutator.() -> Unit)?,
+    spanLimitConfig: SpanLimitConfig
+): SpanLink {
+    val container = AttributesModel(
+        attributeLimit = spanLimitConfig.attributeCountPerLinkLimit,
+        attributeValueLengthLimit = spanLimitConfig.attributeValueLengthLimit
+    )
+    if (attributes != null) {
+        attributes(container)
+    }
```
breedx-splk: Is this more kotlin idiomatic?

```suggestion
    attributes?.invoke(container)
```

breedx-splk (approved):
Thanks for the help, this is great. Much appreciated.

PR #614, opened 2026-06-30
Add app consumer to min-versions test and raise AGP and Gradle floors to 8.0.2

priettt:
## Summary
- Add an app (`com.android.application`) consumer to the `min-versions-android-app` test fixture.
- Raise the minimum supported AGP from **8.0.0 to 8.0.2**.
- Raise the minimum supported Gradle from **8.0 to 8.0.2**.

## Why add an app consumer
The fixture only had a library (`com.android.library`) consumer. Libraries never run D8/R8 — they just package an AAR of JVM bytecode and leave dexing to the app that uses them.

An app consumer fixes that: building it runs D8 over our artifacts (and their transitive dependencies) at the minimum AGP/Gradle/minSdk. That catches two things the library consumer can't:

1. **Toolchain/D8 problems** — like the AGP 8.0.0 crash below.
2. **Dependency conflicts** — e.g. if a dependency we add pulls in two artifacts that ship the same classes. A duplicate-class error only shows up when an app dexes the full dependency graph.

## Why raise the AGP floor to 8.0.2
AGP 8.0.0's bundled D8 crashes when dexing Kotlin 2.0 `@Metadata`. This isn't specific to our library — any app on AGP 8.0.0 with Kotlin 2.x just can't build (unless it manually overrides R8, which is rare). The app consumer catches this; the library consumer can't, because it doesn't dex.


fractalwrench (approved):
Thanks @priettt - can you resolve the conflicts?

fractalwrench:
I'd also personally be in favour of moving to the 8.0.2 patch as our floor.

PR #616, opened 2026-06-30
Add helper action to regenerate yarn lockfile

breedx-splk:
In #606, we needed to do some back-and-forth in order to get the yarn lock to match the updates to the dependencies. Apparently this is not something that renovate will do. One approach is to try and configure renovate to do this as a post-upgrade task, but there are complicated permissions issues with this, and we might not be able to do it.

So I've taken another approach here, that should allow approvers and maintainers to comment in the renovate pr with `/otelbot update-kotlin-js-lock` to trigger a github action that will do the yarn lockfile update and add it to the PR. This will hopefully save us some effort in the long term.

(created in large part with codex)

### DIFF
--- a/.github/workflows/update-kotlin-js-lock.yaml
+++ b/.github/workflows/update-kotlin-js-lock.yaml
@@ -1,0 +1,145 @@
+name: Update Kotlin JS lockfile
+
+# Renovate updates Gradle dependencies, but it does not regenerate the Kotlin/JS
+# Yarn lockfile in kotlin-js-store. Maintainers can use this command on Renovate
+# PRs when dependency updates require lockfile changes by commenting exactly:
+#
+# /otelbot update-kotlin-js-lock
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
+
+jobs:
+  update-kotlin-js-lock:
+    if: >
+      github.event.issue.pull_request &&
+      github.event.comment.body == '/otelbot update-kotlin-js-lock'
+    permissions:
+      contents: read
+      issues: write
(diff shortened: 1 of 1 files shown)


thompson-tomo:
What you could do is eliminate the need to comment and automate it instead based on path filter & branch name. Such an example is https://github.com/open-telemetry/opentelemetry-ruby/blob/main/.github/workflows/renovate-semconv.yml

breedx-splk:
> What you could do is eliminate the need to comment and automate it instead based on path filter & branch name. Such an example is https://github.com/open-telemetry/opentelemetry-ruby/blob/main/.github/workflows/renovate-semconv.yml

Thanks yeah that's true...I'd consider this a baby step in that direction? If we think it's working well but turns out to be a maintainer burden we can proceed with automation. 👍🏻

.github/workflows/update-kotlin-js-lock.yaml
```diff
@@ -0,0 +1,129 @@
...
+#
+# /otelbot update-kotlin-js-lock
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
+
```
Msksgm: The workflow-level `permissions: contents: read` seems redundant because the only job already defines its own permissions block. 
Would it make sense to remove the workflow-level permissions to avoid duplication?

fractalwrench (approved):
LGTM

.github/workflows/update-kotlin-js-lock.yaml
breedx-splk: yeah agreed, let's try that.

PR #623, opened 2026-07-06
Use the new name for the app id / client id in workflows

breedx-splk:
There are warnings, for example over here: https://github.com/open-telemetry/opentelemetry-kotlin/actions/runs/28827882496

I guess the `app-id` key was renamed to `client-id`.


fractalwrench (approved):
LGTM

PR #624, opened 2026-07-07
Add repository settings

### LINKED ISSUE #599: Create repository_settings.md
This is a task for maintainers.

We should be tracking manual settings changes to this repo (outside of terraform), so that it can be more easily recreated if necessary. This also creates a nice audit trail and individual accountability (eg. who changed what when).

See the [android repo for an example](https://github.com/open-telemetry/opentelemetry-android/blob/main/.github/repository-settings.md).

fractalwrench:
## Goal

Closes #599. I believe this is the only change that isn't in terraform already.

### DIFF
--- a/.github/repository_settings.md
+++ b/.github/repository_settings.md
@@ -1,0 +1,15 @@
+# Repository settings
+
+This document describes any changes that have been made to the
+settings for this repository beyond the [OpenTelemetry default repository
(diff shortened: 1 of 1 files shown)


breedx-splk (approved):

breedx-splk:
Technically, I think I created the `protected` environment, but not all the secrets. That can be added in a follow-up.

PR #625, opened 2026-07-07
Implement Tracer enabled()

### LINKED ISSUE #545: Implement enabled() on Tracer
The spec defines an `enabled()` function that determines whether a span should be created: https://opentelemetry.io/docs/specs/otel/trace/api/#enabled

We should add this to the API now that it is stable.

fractalwrench:
## Goal

Closes #545.


breedx-splk (approved):
Looks good to me.

PR #628, opened 2026-07-09
Alter dupe span link behavior

### LINKED ISSUE #547: Allow overriding span links
Adding a link checks if a `SpanContext` already exists and performs no-op. It feels preferable to accept the latest span link and overwrite what is there, as the newest is likely to be more desired for attributes/trace context: https://github.com/open-telemetry/opentelemetry-kotlin/blob/aacfe939ba3d22d432e2b445c5415dfa6d6e2d71/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt#L128

fractalwrench:
## Goal

Closes #547 by allowing span links to be added as long as the link limit is not exceeded. The existing approach is overly restrictive and there's not anything in the specification that prevents duplicate links being added.


bidetofevil (approved):
LGTM

PR #629, opened 2026-07-09
Add counts for dropped events

### LINKED ISSUE #544: Report dropped count for attributes/events/links
The specification states that the count of dropped attributes, events, and links should be recorded in OTLP: https://opentelemetry.io/docs/specs/otel/common/mapping-to-non-otlp/#dropped-attributes-count

Currently this information is not tracked so the value is always set to 0 in OTLP. We should fix that in the implementation module.

fractalwrench:
## Goal

Adds a count for dropped events as required by the specification. This partially resolves #544


bidetofevil (approved):
LGTM

breedx-splk (approved):

PR #631, opened 2026-07-09
Record dropped attribute count

### LINKED ISSUE #544: Report dropped count for attributes/events/links

fractalwrench:
## Goal

Adds a count for dropped attributes as required by the specification. This partially resolves #544


bidetofevil (approved):
LGTM

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesModel.kt
```diff
@@ -138,11 +138,32 @@ internal class AttributesModel(
...
+            return false
+        }
+        if (attrs.contains(key)) {
+            // Overwriting an existing attribute never drops.
+            return true
+        }
+        if (attrs.size < attributeLimit) {
+            return true
+        }
+        droppedAttributesCountImpl++
```
breedx-splk: I understand why it is convenient to put this here, but it feels misplaced. To the casual reader, simply checking to see if an attribute _can_ be added has the potential side effect of incrementing the number of dropped attributes. It's not immediately clear that the check and the set are coupled, and bug could decouple them.

One way of improving this would be to make this method take a lambda, so it looks more like this:

```kotlin
private fun ifPreconditionsOk(key: String, setter: () -> Unit) {
        if (key.isEmpty()) {
            // Invalid key: ignored, not counted as a dropped attribute.
            return
        }
        if (attrs.contains(key)) {
            // Overwriting an existing attribute never drops.
            return 
        }
        if (attrs.size >= attributeLimit) {
            droppedAttributesCountImpl++
            return 
        }
        return setter()
    }
```

and then the call sites look more like
```kotlin
ifPreconditionsOk(key) {
    attrs[key] = value
}
```

breedx-splk (approved):
It looks good to me, nothing too blocking, but I did have one comment that I think is worth considering. Thanks!

PR #633, opened 2026-07-09
Add klib consumer to min-versions test and document Kotlin floors per target

priettt:
## Summary
- Add a `min-versions-kmp-app` fixture that consumes the published iOS/JS klibs, enforcing the Kotlin floor for klib consumers.
- Document the per-target Kotlin floors in the changelog: **2.0.0 for JVM/Android** consumers, while **iOS/JS consumers need at least the Kotlin version this library is built with**.
- Consolidate the integration tests into a macOS CI job.

## Why klib consumers have a higher floor
Klibs are stamped with `abi_version` = the build Kotlin's version, and older compilers reject them outright. This is a klib toolchain constraint shared by all KMP libraries — not something this library can lower.

https://youtrack.jetbrains.com/issue/KT-76131

## CI changes
Integration tests now run in a single `integration-test` job on `macos-latest`, because only a macOS host can publish (and therefore consume) the Apple klibs.

## Testing
Ran the exact CI sequence locally on macOS (`publishToMavenLocal` + `:gradle-integration-test:test`): both fixtures pass — the JVM/Android one at Gradle/AGP 8.0.2 and Kotlin 2.0.0, the klib one at Kotlin 2.4.0 with iOS + JS targets.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -38,12 +38,6 @@ jobs:
       - name: Gradle Build
         run: ./gradlew build :benchmark-android:assemble :benchmark-android:compileReleaseAndroidTestKotlin :benchmark-jvm:assemble :examples:example-app-android:assembleRelease koverXmlReport -x :gradle-integration-test:test --stacktrace
 
-      - name: Publish snapshot to Maven local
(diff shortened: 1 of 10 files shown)


fractalwrench (approved):
LGTM - thanks!

PR #634, opened 2026-07-09
[DNM] Demo PR to use the federated semantic conventions

bidetofevil:
Demonstration of how we can depend on a federated semantic conventions repo. The relevant changes are where we are fetching our dependent registry from which we generate the Kotlin files, and a command line parameter to generate the constant files for the dependent registries that are transitively pulled in (i.e. the core otel semconv)


fractalwrench (commented):
This is cool. My one comment is that because opentelemetry-kotlin runs on the client only for certain targets, we probably want to create a `semconv-client` module that is a home for client-specific values.

bidetofevil:
Yeah, there are a couple of ways we can organize how consumable artifacts are generated. 

Right now the project is like an analog to `semantic-conventions-java` in that it generates artifacts for the core registry in Kotlin. But now that we have a new registry that extends it, if we did the same thing (i.e. what this PR does), we are effectively creating Kotlin artifacts for a different registry - and there's generated artifacts for the core registry. 

And whereas in the past iOS, JS, etc. can use the artifacts for the core registry built for other languages, they can't do that with the end user client registry. 

So should we generate it for all the support platforms? Possibly. That would be a reasonable first step. But then there's also the matter of versioning. Right now if the artifacts are versioned with the SDK/API, but their real version is what is the new repo. So should we consider spinning that off in some ways then? Is this another repo?

Anyway, lots to discuss after the initial repo is formally created

bidetofevil:
Per the discussion in the SIG, I'm going modify this to create `semconv-client` like Jamie suggested just to see what it's like, rather than replace the existing project that builds the constant files for the core OTel semconv. 

Whether we actually build these artifacts it in the long term depends on the plan for the federated repo going forward, i.e. if that repo isn't going to build it and tag it with that repo's version, SOMEONE has to, and this could be that someone.

bidetofevil:
@priettt posted a discussion on this already regarding support in iOS/JS: https://github.com/Kotlin/kotlinx.coroutines/issues/3326

fractalwrench:
Closing this as I don't think it's actively being worked on - feel free to reopen if needed.

PR #635, opened 2026-07-10
Alter OtlpClient to handle partial success response

### LINKED ISSUE #558: Avoid relying on 200 status code to indicate OTLP export success
Currently the OTLP exporters assume that 200 status code is a complete success. This can be a partial success in which case the response body may need to be read: https://opentelemetry.io/docs/specs/otlp/#otlpgrpc-response

fractalwrench:
## Goal

Alters `OtlpClient` so that it handles a partial success response even when the status code is 200. I've additionally hardened the client so that it doesn't crash when a malformed body is returned.

Closes #558

## Testing

Added test cases.


lemnik (approved):
LGTM

bidetofevil (approved):
LGTM

PR #638, opened 2026-07-10
`OtelJavaSpanBuilderAdapter.setStartTimestamp` should obey TimeUnit

lemnik:
## Goal
`OtelJavaSpanBuilderAdapter.setStartTimestamp` should convert the given `TimeUnit` into nanoseconds instead of directly copying the `startTimestamp` with no conversion

## Testing
Updated a new unit test to `OtelJavaSpanExportTest` inline with the existing tests for the `OtelJavaSpanBuilderAdapter`


fractalwrench (approved):
LGTM

PR #644, opened 2026-07-13
Handle failure cases in exporters

### LINKED ISSUE #639: OTLP exporter can crash the app on network failure (only HttpRequestTimeoutException is caught; export runs on a scope without a CoroutineExceptionHandler)
### Describe the bug

When an OTLP HTTP export fails with anything other than a timeout (e.g. `ConnectException` — device offline, connection refused, host unreachable), the exception becomes an **uncaught coroutine exception** and crashes the host app. Telemetry export should be best-effort and must never crash the app (related to #502 — this is the exporter path, which that issue didn't cover).

**Root cause (in common code):**

1. `OtlpClient.exportTelemetry` (`commonMain/io/opentelemetry/kotlin/export/OtlpClient.kt`) wraps the request in `try { httpClient.post(...) … } catch (ignored: HttpRequestTimeoutException) { Unknown }` — it catches **only** `HttpRequestTimeoutException`. A `ConnectException` (or any other `IOException`) propagates out.
2. `TelemetryExporter` (`commonMain/io/opentelemetry/kotlin/export/TelemetryExporter.kt`) runs the export fire-and-forget: `private val scope = CoroutineScope(SupervisorJob())` and `export()` does `scope.launch { exportTelemetry(...) }`. The scope has **no `CoroutineExceptionHandler`**, so the propagated exception reaches the default uncaught handler → the app crashes.

This affects both `SimpleSpanProcessor` and `BatchSpanProcessor` (both export via `OtlpHttpSpanExporter` → `TelemetryExporter`).

### Steps to reproduce

1. Configure an OTLP/HTTP span exporter pointing at an unreachable endpoint (or put the device offline / airplane mode).
2. Create and end a span so it gets exported.
3. The export coroutine throws `ConnectException`, which is uncaught → app crash.

### Environment
* OS version: Android (reproduced on emulator); cause is in common code, so other targets are affected too
* SDK version: 0.4.0

### Example code snippet 

```kotlin
val otel = createOpenTelemetry {
    serviceName = "demo"
    tracerProvider {
        export { simpleSpanProcessor(otlpHttpSpanExporter(baseUrl = "http://10.0.2.2:4318")) }
    }
}

// With no collector reachable (device offline):
val tracer = otel.getTracer("demo")
tracer.startSpan("test").end() // export fails with ConnectException -> uncaught -> crash
```

<details><summary>Error messages:</summary>

```
FATAL EXCEPTION: DefaultDispatcher-worker-9
java.net.ConnectException: Failed to connect to /10.0.2.2:4318
    at okhttp3.internal.connection.ConnectPlan.connectSocket(ConnectPlan.kt:278)
    at okhttp3.internal.connection.ConnectPlan.connectTcp(ConnectPlan.kt:135)
    at okhttp3.internal.connection.FastFallbackExchangeFinder$launchTcpConnect$1.runOnce(FastFallbackExchangeFinder.kt:141)
    at okhttp3.internal.concurrent.TaskRunner$runnable$1.run(TaskRunner.kt:81)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1156)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:651)
    at java.lang.Thread.run(Thread.java:1119)
    Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}, Dispatchers.Default]
Caused by: java.net.ConnectException: failed to connect to /10.0.2.2 (port 4318) ... ENETUNREACH (Network is unreachable)
```
</details>

### Expected behaviour

Export failures are swallowed/retried; the host app never crashes.

### Suggested fix (either / both)

- In `OtlpClient.exportTelemetry`, catch general failures (e.g. `IOException` / any non-`CancellationException`) around the HTTP call and map them to `OtlpResponse.Unknown`, not just `HttpRequestTimeoutException` (re-throwing `CancellationException`).
- Give `TelemetryExporter`'s `scope` a `CoroutineExceptionHandler` (and/or wrap the `exportTelemetry` body in try/catch) so a thrown export can never become an uncaught coroutine exception.

### Workaround

Wrapping the Ktor engine passed to `otlpHttpSpanExporter(baseUrl, httpClientEngine = …)` so transport failures are converted to a `503` response (extending `HttpClientEngineBase` so `install()` binds the wrapper's `execute()`), which the exporter then treats as a retryable server error — keeping export best-effort.

fractalwrench:
## Goal

Closes #639 by better handling failure cases in a few exporters.

### DIFF
--- a/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/export/BatchTelemetryProcessor.kt
+++ b/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/export/BatchTelemetryProcessor.kt
@@ -18,7 +18,8 @@ internal class BatchTelemetryProcessor<T>(
 ) : TelemetryCloseable {
 
     private val shutdownState: MutableShutdownState = MutableShutdownState()
-    private val scope = CoroutineScope(SupervisorJob() + dispatcher)
+    private val scope =
+        CoroutineScope(SupervisorJob() + dispatcher + telemetryExceptionHandler("Batch processor"))
     private val mutex = Mutex()
     private val queue = mutableListOf<T>()
 
--- a/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/export/CoreLogRecordExporterApi.kt
+++ b/exporters-core/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/export/CoreLogRecordExporterApi.kt
@@ -4,6 +4,7 @@ package io.opentelemetry.kotlin.logging.export
 import io.opentelemetry.kotlin.ExperimentalApi
 import io.opentelemetry.kotlin.error.NoopSdkErrorHandler
 import io.opentelemetry.kotlin.export.BatchTelemetryDefaults
+import io.opentelemetry.kotlin.export.telemetryExceptionHandler
 import io.opentelemetry.kotlin.init.LogExportConfigDsl
 import io.opentelemetry.kotlin.platformLog
 import kotlinx.coroutines.CoroutineDispatcher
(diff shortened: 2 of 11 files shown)


lemnik (approved):
LGTM

bidetofevil (approved):
LGTM

PR #647, opened 2026-07-13
Pin the klib consumer Kotlin floor at 2.4.0

priettt:
## Goal

The min-versions klib fixture consumed the published iOS/JS artifacts with whatever Kotlin this library is currently built with (the `test-kotlin-multiplatform-klib` alias tracked the `kotlin` catalog version), so every Kotlin bump would silently raise the floor for klib consumers.

Pin the floor in the version catalog instead (`minSupportedKotlinKlib = 2.4.0`). A Kotlin update that outgrows the pin now fails the min-versions integration test, making a floor raise a deliberate, documented decision rather than a side effect of a dependency update.

The new pin is also added to Renovate's never-bump list (#645), matching the other minimum-supported version pins.

## Testing

Ran `publishToMavenLocal -PsnapshotPublish=true` followed by `:gradle-integration-test:test` locally (macOS): both fixtures pass, with the klib fixture resolving KGP 2.4.0 through the new pin.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -18,7 +18,7 @@
       "description": "Never bump the minimum-supported version pins in the version catalog. sharedVariableName is the version.ref of the catalog entry, with '-'/'_' normalized to '.'",
       "matchManagers": ["gradle"],
       "matchJsonata": [
-        "sharedVariableName in ['android.minSdk', 'android.minCompileSdk', 'android.minAgpVersion', 'minSupportedKotlinLang', 'minSupportedKotlinPlugin', 'minSupportedGradle']"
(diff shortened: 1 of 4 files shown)


fractalwrench (commented):
Does this prevent running `MinSupportedVersionTest` against Kotlin 2.0.0 on the JVM? Or does this just apply to iOS/JS?

fractalwrench:
Created #648 which tracks the issue

priettt:
> Does this prevent running `MinSupportedVersionTest` against Kotlin 2.0.0 on the JVM? Or does this just apply to iOS/JS?

@fractalwrench it doesn't, Kotlin 2.0 is still the min version supported for JVM, and it has its own test.

fractalwrench (approved):
LGTM

PR #656, opened 2026-07-14
Setup yaml parser

### LINKED ISSUE #535: Add yaml parser for declarative config
Add a YAML parser that is capable of reading a YAML file for declarative config. Ideally this 3rd party library should support KMP and optionally be usable with kotlinx.serialization as we're already using that. This issue can be considered closed when the infrastructure is in place to read an empty YAML file and return _some_ data model object (which doesn't necessarily need to match the objects in #533 as an initial step)

https://opentelemetry.io/docs/specs/otel/configuration/data-model/#yaml-file-format

May affect #534, as we may choose to isolate this in an opt-in module such as `opentelemetry-config-yaml` so that a yaml dependency isn't forced on end-users by default.

fractalwrench:
## Goal

Closes #535 by setting up a YAML parser that can be used for declarative configuration. https://github.com/him188/yamlkt seems to be the best choice for KMP YAML parsers as it's used by Ktor.

## Testing

Added a unit test.

### DIFF
--- a/config/build.gradle.kts
+++ b/config/build.gradle.kts
@@ -12,6 +12,19 @@ kotlin {
         val commonMain by getting {
             dependencies {
                 implementation(project(":sdk-api"))
+                implementation(libs.yamlkt)
+                implementation(libs.okio)
+            }
+        }
+        val commonTest by getting {
+            dependencies {
+                implementation(libs.kotlin.test)
+                implementation(libs.okio.fakefilesystem)
+            }
+        }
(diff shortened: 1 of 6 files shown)


lemnik (approved):
LGTM

bidetofevil (approved):
LGTM. That library has the right license etc. and safe for us to use?

fractalwrench:
yamlkt is licensed with Apache 2.0 which is the same as this repo

Issue #44, opened 2026-01-15
Verify that all OTLP HTTP exporters support spec requirements


fractalwrench:
Verify that all OTLP HTTP exporters support spec requirements listed here: https://opentelemetry.io/docs/specs/otlp/

- [ ] Concurrent sending
- [x] Retryable responses with backoff
- [x] Handles non-retryable responses
- [ ] Honors throttling response
- [ ] Multi-destination spec-compliance

PR #660, opened 2026-07-14
Alter telemetry export to comply with spec

fractalwrench:
## Goal

Works towards addressing #44 by implementing exponential backoff and adding sufficient test coverage.


bidetofevil (approved):
TBH, I'm not sure how impactful this will be given API end points that mobile apps connect to should already be capable of dealing with thundering herds. The fact that this might cause telemetry to be sent out of order - is that a concern?

If it's part of the spec, sure, but I really don't know if this applies to us.

fractalwrench:
This is part of the [OTLP specification](https://opentelemetry.io/docs/specs/otlp/#otlphttp-connection). Additionally, this library could be used on the backend where this is more of a problem.

PR #666, opened 2026-07-15
Create Kotlin classes from opentelemetry-config schema

### LINKED ISSUE #533: Generate code from opentelemetry-configuration data models
We should create a CI action that generates Kotlin data classes from the JSON schema in opentelemetry-configuration that defines model objects. This CI action should run gradle task(s) that downloads the latest JSON schema, generates source code, commits the source code, and opens a PR. There is a similar CI action for generating semantic conventions that could be used as rough inspiration.

https://opentelemetry.io/docs/specs/otel/configuration/data-model/#overview
https://github.com/open-telemetry/opentelemetry-configuration

fractalwrench:
## Goal

Closes #533 by generating Kotlin classes in a build-time task from [opentelemetry-configuration's schema](http://github.com/open-telemetry/opentelemetry-configuration).

 [KotlinPoet](https://square.github.io/kotlinpoet/) was used to generate the code. A GH workflow exists that will create a PR with changes to the generated code when a new version of opentelemetry-configuration is released.

I created a new model named `config-schema` as I think it probably makes sense to separate the model classes from parsing of YAML. We could also use these model classes in the config DSL rather than defining custom classes like we do now.

Most of this changeset is generated code - the important parts to review are the build task and model generator.


bidetofevil (commented):
Do we want to check in the generated files or can we consume the built artifact instead? @breedx-splk probably as an opinion about this

fractalwrench:
I'd lean towards checking in files for this (and for semconv) as it avoids the requirement to download the schema and run codegen each build. That step should be cached, but it's one more thing that can go wrong...

buildSrc/src/main/kotlin/io/opentelemetry/kotlin/config/schema/ConfigModelGenerator.kt
```diff
@@ -0,0 +1,214 @@
...
+        return oneOf.any { asMap(it)["type"] == "null" }
+    }
+
+    private fun kdoc(schema: Map<String, Any?>): String? {
+        val description = (schema["description"] as? String)?.trim().orEmpty()
+        if (description.isEmpty()) {
+            return null
+        }
+        return description.replace("*/", "* /")
+    }
```
lemnik: Would this possible read better as a `Documentable` extension function? It would reduce:

```kotlin
.apply { kdoc(def)?.let { addKdoc("%L", it) } }
```
to:
```kotlin
.deriveKDoc(def)
```

```suggestion
    private fun <R: Documentable> R.deriveKDoc(schema: Map<String, Any?>): R {
        val description = (schema["description"] as? String)?.trim().orEmpty()
        if (description.isEmpty()) {
            return this
        }
       addKDoc("%L", description.replace("*/", "* /"))
       
       return this
    }
```

bidetofevil (approved):
LGTM

PR #667, opened 2026-07-15
Supply root context to ImplicitContextStorage

fractalwrench:
## Goal

Alters the API so that if a custom `ImplicitContextStorage` is specified, a root context supplier is available. This is important as custom storage will always want to start with root context as the default, but currently it's not possible to access this until after the SDK has initialized.


bidetofevil (approved):
LGTM. API changes are OK since we're not stable, eh

PR #677, opened 2026-07-15
Prevent invalid span names

### LINKED ISSUE #668: Disallow empty string as span name
`Tracer` currently allows empty strings as a valid span name. We should disallow this and return a `NonRecordingSpan` if an end-user attempts this.

fractalwrench:
## Goal

Closes #668 by disallowing empty or blank strings as span names. Surprisingly the spec doesn't disallow empty strings as span names, but I can't imagine this is a useful behavior...

## Testing

Added unit tests.

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt
@@ -51,6 +51,10 @@ internal class TracerImpl(
         action: (SpanCreationAction.() -> Unit)?
     ): Span =
         shutdownState.ifActiveOrElse(noopSpan) {
+            if (name.isBlank()) {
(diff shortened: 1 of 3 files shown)


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/TracerImpl.kt
```diff
@@ -51,6 +51,10 @@ internal class TracerImpl(
         action: (SpanCreationAction.() -> Unit)?
     ): Span =
         shutdownState.ifActiveOrElse(noopSpan) {
+            if (name.isBlank()) {
+                return@ifActiveOrElse NonRecordingSpan(invalidSpanContext, invalidSpanContext)
```
breedx-splk, line 55: this stupid syntax gets me every time. 🤣

breedx-splk (approved):

PR #678, opened 2026-07-15
Tweak default schedule delay

### LINKED ISSUE #674: Default batch log record processor schedule delay to 1000ms
Currently it defaults to 5000ms but the spec states 1000ms should be used: https://opentelemetry.io/docs/specs/otel/logs/sdk/#batching-processor

fractalwrench:
## Goal

Closes #674 by using the correct default scheduling delay for batch log record processors.


bidetofevil (approved):
LGTM in terms of the code change. But 5s seems an awful long time on mobile. While it says that on the spec, is that really a reasonble default?

fractalwrench:
IMO 5s is reasonable for an export interval. We could always assess the defaults separately?

PR #679, opened 2026-07-15
Extract ThreadLocal code

fractalwrench:
## Goal

The KMP `expect/actual` implementation of `ThreadLocal` was tied in with `ThreadLocalImplicitContextStorage`. I've moved it to the `platform-implementations` module so that it can be used by other code that requires thread-local, and also so that the context implementation is simpler.

## Testing

Added unit tests.


platform-implementations/src/jsMain/kotlin/io/opentelemetry/kotlin/ThreadLocal.js.kt
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.kotlin
+
+/**
+ * JavaScript implementation of [ThreadLocal]. The runtime is single-threaded, so this is a single
+ * shared slot.
+ */
+public actual class ThreadLocal<T> actual constructor() {
```
bidetofevil, line 12: Are web workers running on the same thread too?

bidetofevil (approved):
LGTM. I have a question about JS but that's largely due to my ignorance of their threading model wrt web workers. Maybe that's not even native JS but just the browser runtime, in which case you can ignore my comment.

platform-implementations/src/jsMain/kotlin/io/opentelemetry/kotlin/ThreadLocal.js.kt
fractalwrench: This implementation doesn't handle web workers/node. I figure if that becomes a problem that people care about, we can address it at that point.

PR #680, opened 2026-07-15
Allow supplying custom IdGenerator

### LINKED ISSUE #670: Provide mechanism for setting custom IdGenerator
The spec states that a mechanism for altering trace/span ID generation should be provided: https://opentelemetry.io/docs/specs/otel/trace/sdk/#id-generators

IMO this should be done on the `OpenTelemetryConfigDsl` (using similar syntax to the approach of `ImplicitContextStorage`)

fractalwrench:
## Goal

Closes #670 by allowing a custom `IdGenerator` to be supplied as mandated by the spec.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #684, opened 2026-07-16
Provide API to enhance implicit context use in coroutines

### LINKED ISSUE #478: Creates a coroutine-based approach for storing the implicit context
Creates a coroutine-basedapproach for storing the implicit context

fractalwrench:
## Goal

Closes #478 by providing an API that can be used to manage implicit context from coroutines. This API has the caveat that it's only available for the JVM target due to a restriction in the API of kotlinx.coroutines where [ThreadContextElement](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-thread-context-element/) is only available on the JVM. Consequently this is provided as an opt-in module for users who want this functionality.

Usage would look like this:

```
withContext(Dispatchers.Default + myOtelContext.asCoroutineContextElement()) {
     // perform operation
 }
```

`withContext` is the kotlinx.coroutine API to launch a coroutine. Along with the dispatcher, opentelemetry-kotlin's `Context` object is converted to `CoroutineContextElement` via `asCoroutineContextElement()`. Under the hood, this will be invoked by the coroutine when the coroutine is suspended and restored, which allows for the context to be tracked appropriately within the coroutine.

There are limitations about how automatic it's possible to make this due to the existing APIs. IMO this is probably one of the best, less invasive approaches we could take.

## Testing

Added unit tests.

### DIFF
--- a/context-coroutines/api/jvm/context-coroutines.api
+++ b/context-coroutines/api/jvm/context-coroutines.api
@@ -1,0 +1,4 @@
+public final class io/opentelemetry/kotlin/context/CoroutineContextElementKt {
+	public static final fun asCoroutineContextElement (Lio/opentelemetry/kotlin/context/Context;)Lkotlin/coroutines/CoroutineContext$Element;
+}
+
--- a/context-coroutines/build.gradle.kts
+++ b/context-coroutines/build.gradle.kts
@@ -1,0 +1,25 @@
+plugins {
+    kotlin("multiplatform")
+    id("com.android.kotlin.multiplatform.library")
+    id("io.opentelemetry.kotlin.build-logic")
+    id("signing")
+    id("com.vanniktech.maven.publish")
+    id("org.jetbrains.kotlinx.kover")
+}
+
+kotlin {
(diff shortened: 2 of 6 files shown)


lemnik (approved):

bidetofevil (approved):
LGTM. So the existing extension in OTel Java that works with Coroutines has the same limitation? And it means non-JVM targets can't have automatic context propgatation via Coroutine Scope?

fractalwrench:
Non-JVM targets don't get to use this API. `ThreadLocalContextElement` isn't available on non-JVM platforms.

opentelemetry-java's approach to coroutine context management uses a similar implementation:

https://github.com/open-telemetry/opentelemetry-java/blob/main/extensions/kotlin/src/main/kotlin/io/opentelemetry/extension/kotlin/ContextExtensions.kt
https://github.com/open-telemetry/opentelemetry-java/blob/main/extensions/kotlin/src/main/java/io/opentelemetry/extension/kotlin/KotlinContextElement.java

PR #685, opened 2026-07-16
Make Carrier argument nullable on Propagator API

### LINKED ISSUE #665: Adjust TextMapPropagator signature to match spec and TextMapGetter/TextMapSetter
While working on the B3 propagator (#605 ) I noticed an inconsistency. 
After #595  was merged, the carrier on TextMapSetter.set / TextMapGetter.get is nullable (`C?`), but TextMapPropagator.inject / extract still take a non-null `carrier: T` ([code](https://github.com/jdvr/opentelemetry-kotlin/blob/main/api/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TextMapPropagator.kt#L26)).

So the propagator promises a non-null carrier while the setter/getter it delegates to accept null. It's also worth noting opentelemetry-java declares the carrier @Nullable on inject/extract.

We should make TextMapPropagator's carrier nullable too, for consistency with the getter/setter and the spec

fractalwrench:
## Goal

Closes #665 by making the `Carrier` argument nullable.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -23,6 +23,7 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Setup Java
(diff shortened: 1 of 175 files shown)


bidetofevil (approved):
LGTM

PR #687, opened 2026-07-17
OTLP File Exporter and JSON Encoding Implementation 

viorel-alexandrescu:

### DIFF
--- a/exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/encode/OtlpJsonEncoder.kt
+++ b/exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/encode/OtlpJsonEncoder.kt
@@ -1,0 +1,5 @@
+package io.opentelemetry.kotlin.encode
+
+internal interface OtlpJsonEncoder<in T> {
+    fun encode(value: T): String
+}
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -39,6 +39,7 @@ include(
     ":exporters-otlp",
     ":exporters-persistence",
     ":exporters-protobuf",
+    ":exporters-json",
     ":java-typealiases",
     "examples:example-app",
     "examples:example-app-android",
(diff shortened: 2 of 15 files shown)


settings.gradle.kts
```diff
@@ -40,11 +39,11 @@ include(
     ":exporters-otlp",
     ":exporters-persistence",
     ":exporters-protobuf",
+    ":exporters-json",
```
fractalwrench, line 42: This module feels like a good choice.

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/encode/OtlpJsonEncoder.kt
```diff
@@ -0,0 +1,5 @@
+package io.opentelemetry.kotlin.encode
+
+internal interface OtlpJsonEncoder<in T> {
+    fun encode(value: T): String
```
fractalwrench, line 4: Is there a way of buffering this encoding so that encoding happens on a stream, rather than a whole string? That will reduce memory requirements for larger payloads

fractalwrench (commented):
This broadly looks like it's taking the right approach. If you were looking for further ways to split it up, one PR could contain the JSON model classes, and the other could contain all the exporters.

viorel-alexandrescu:
> This broadly looks like it's taking the right approach. If you were looking for further ways to split it up, one PR could contain the JSON model classes, and the other could contain all the exporters.

Thanks! That sounds like a plan to me, I'll split these changes up to better track their respective functionalities.

We can pick up on the details of each implement after I raise those PRs.

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/encode/OtlpJsonEncoder.kt
viorel-alexandrescu: I'll tackle it in [this PR](https://github.com/open-telemetry/opentelemetry-kotlin/pull/688)

PR #688, opened 2026-07-17
JSON Encoding / Export for LogRecord and SpanData

viorel-alexandrescu:
## Goal

Implement JSON Encoding / Export of LogRecord and SpanData entities


fractalwrench (commented):
Thanks for opening this @viorel-alexandrescu. Can you address the failing CI checks and some tests please? The code itself looked reasonable but I'd like confidence it's passing automated tests before reviewing further.

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/export/JsonLogRecordExporterApi.kt
```diff
@@ -0,0 +1,13 @@
+
+package io.opentelemetry.kotlin.logging.export
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.init.LogExportConfigDsl
+
+/**
+ * Creates an in-memory log record exporter that stores telemetry in memory.
+ * This is intended for development/testing rather than production use.
```
fractalwrench: Is this comment relevant for `jsonLogRecordExporter()` or is it carried over from one of the other functions?

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/JsonSpanExporterApi.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.init.TraceExportConfigDsl
+
+/**
+ * Creates an in-memory log record exporter that stores telemetry in memory.
+ * This is intended for development/testing rather than production use.
+ */
+@ExperimentalApi
+public fun TraceExportConfigDsl.jsonSpanExporter(): JsonSpanExporter =
```
fractalwrench: Is this comment relevant for `jsonSpanExporter()` or is it carried over from one of the other functions?

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/encode/OtlpJsonEncoder.kt
```diff
@@ -0,0 +1,18 @@
+package io.opentelemetry.kotlin.encode
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.json.Json
+
+sealed interface OtlpJsonEncoder<in T, S> {
+    fun getSerializable(value: T): S
+    fun getSerializer(): KSerializer<S>
+    fun encode(value: T): Sequence<String> =
```
fractalwrench: Is there a reason a sequence of strings is returned here rather than an InputStream/equivalent?

exporters-json/src/commonTest/kotlin/io/opentelemetry/kotlin/encode/JsonLogRecordEncoderTest.kt
```diff
@@ -0,0 +1,20 @@
...
+    @Test
+    fun `should successfully encode a log record data in JSON format`() {
+        // given
+        val encoder = JsonLogRecordEncoder()
+
+        // when
+        val result = encoder.encode(FakeReadableLogRecord())
+
+        // then
+        assertTrue { result.any() }
```
fractalwrench: I think it'd be sensible to add a unit test either here or in the OTLP exporters that assert _some_ JSON is actually serialized for both logs and spans. Right now the test coverage is fairly non-specific - we just know that _some_ object was created

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/export/JsonLogRecordExporterApi.kt
viorel-alexandrescu: It was carried over, my mistake.
But if you have any suggestion, I'm open to use that.

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/export/JsonSpanExporterApi.kt
viorel-alexandrescu: It was carried over, I can delete it if it's unnecessary.

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/encode/OtlpJsonEncoder.kt
viorel-alexandrescu: This was carried over from the initial set of changes

exporters-json/src/commonTest/kotlin/io/opentelemetry/kotlin/encode/JsonLogRecordEncoderTest.kt
viorel-alexandrescu: Good shout! I'll add one

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/encode/OtlpJsonEncoder.kt
fractalwrench: IMO it'd be preferable to encode this to an `InputStream` for flexibility. E.g. `fun encode(value: T, stream: InputStream): Sequence<String>`

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/export/JsonLogRecordExporterApi.kt
fractalwrench: 'Exports log records as OTLP JSON' would be sufficient for this and the `jsonSpanExporter` function

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/encode/OtlpJsonEncoder.kt
viorel-alexandrescu: I haven't found a viable option to use for the `InputStream` parameter, do you have any suggestions?
Ktor has an [article](https://ktor.io/docs/io-interoperability.html) on this, but I'm not sure we should use that.

fractalwrench: This artifact adds Okio bindings to kotlinx-serialization: https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json-okio/kotlinx.serialization.json.okio/

Then it should be possible to use Sinks and Buffers, which will have the same effect as using streams. It's preferable to use this sort of I/O as it avoids having to write an entire JSON string into memory

viorel-alexandrescu: So in this case, how would the function signature look like exactly? I think I understood your suggestion, but I'm confused if we need a sequence anymore.

fractalwrench: Something along these lines:

```kotlin
    fun encode(value: T, sink: BufferedSink): Unit
```

viorel-alexandrescu:
I think we've combined two concerns within this PR.

You mentioned the encoding function signature could be something like `fun encode(value: T, sink: BufferedSink)`.

My initial thought process was to have a separate class to handle this part (writing to a file), my approach being broadly expressed in [PR #689](https://github.com/open-telemetry/opentelemetry-kotlin/pull/689).

Your argument about not keeping the whole JSON in memory is quite sound, but I think we should not haste this optimization yet. I do believe it will be kept in memory regardless, as long as we have a reference to the serialization result.

I propose a different approach to this, let the encoder just be an abstraction over what serialization mechanism we use and have a dedicated class to write to the file (as I suppose is the approach for other formats).

And if we do need to optimize how these results are passed between these components, we could use a `Sequence` or something similar.

What are your thoughts on this @fractalwrench ?

exporters-json/src/commonMain/kotlin/io/opentelemetry/kotlin/encode/OtlpJsonEncoder.kt
```diff
@@ -0,0 +1,18 @@
...
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.json.Json
+import okio.BufferedSink
+
+sealed interface OtlpJsonEncoder<in T, S> {
+    fun getSerializable(value: T): S
+    fun getSerializer(): KSerializer<S>
+    fun encode(value: T, sink: BufferedSink) {
+        sink.writeUtf8(
+            Json.encodeToString(
```
fractalwrench: FYI: this is the API that allows serializing JSON to a stream: https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/encode-to-stream.html

fractalwrench (commented):
@viorel-alexandrescu agree that I may have been hasty in optimizing this, particularly if the `encodeToStream` API is currently `ExperimentalApi` in kotlinx.serialization.

I'd suggest fixing the checks up and then I'd be happy to approve - future work can happen in future PRs.

viorel-alexandrescu:
I have found in the logs that the tests fail due to this issue:

```
Required by:

        project : > io.opentelemetry.kotlin:exporters-json:0.7.0-SNAPSHOT > io.opentelemetry.kotlin:exporters-json-iosarm64:0.7.0-SNAPSHOT



    * Get more help at 
    https://help.gradle.org/


    BUILD FAILED
     in 36s

    Watched directory hierarchies: [/private/var/folders/df/djsxfhc17x95674wsm_g8s980000gn/T/junit-13563262705557936040]


MinSupportedVersionsTest > min supported versions can consume klib artifacts(Path) FAILED
    org.gradle.testkit.runner.UnexpectedBuildFailure at MinSupportedVersionsTest.kt:57

2 tests completed, 2 failed
```

@fractalwrench I've looked into `MinSupportedVersionsTest` but I'm not sure what I need to amend to fix it. Could you help me with some input?

fractalwrench:
@viorel-alexandrescu it looks like `integration-test` has been added as an `implementation` dependency in the `build.gradle` of the new otlp-json module, which fails as it's adding a test module as a dependency of a production module.

viorel-alexandrescu:
> @viorel-alexandrescu it looks like `integration-test` has been added as an `implementation` dependency in the `build.gradle` of the new otlp-json module, which fails as it's adding a test module as a dependency of a production module.

True, but that's where I'm referencing `SerializableLogRecordData` and `SerializableSpanData` from.
Should I move those models to a shared module?

fractalwrench:
@viorel-alexandrescu yes please.

viorel-alexandrescu:
> @viorel-alexandrescu yes please.

Okay, where would it be best for them to be copied to?

fractalwrench:
@viorel-alexandrescu I'd suggest creating a new `serializable-models` module that contains those modules and the test coverage. It's probably best attempting that in a separate PR then rebasing this PR on top of that.

viorel-alexandrescu:
> @viorel-alexandrescu I'd suggest creating a new `serializable-models` module that contains those modules and the test coverage. It's probably best attempting that in a separate PR then rebasing this PR on top of that.

Done. You can have a look over the PR #1011

viorel-alexandrescu:
> This has been automatically marked as stale because it has been marked as needing author feedback and has not had any activity for 14 days. It will be closed automatically if there is no response from the author within 14 additional days from this comment.

Soon will merge #1011 and close this as well

fractalwrench:
#1011 has merged so this could use a merge/rebase with main

Issue #378, opened 2026-04-07
Create Instrument API


fractalwrench:
Create the Instrument API as described in the spec: https://opentelemetry.io/docs/specs/otel/metrics/api/

fractalwrench:
@inetand I can't seem to assign this to you in the GitHub UI, but feel free to take this on.

inetand:
@fractalwrench OK, thanks for letting me know! I'll work on it.

fractalwrench:
It seems like it needed some interaction with the issue first. I've assigned you now

colintheshots:
@inetand It seems a number of issues are blocked on this one. What still remains after #695? Is there anywhere I can help?

inetand:
@colintheshots Apologies for the lack of updates, I was on vacation for a while.

That is a fair question. This issue is quite broad because it covers most of the foundation required for concrete instrument implementations. Maybe I should have created and linked smaller prerequisite issues first, such as [#938](https://github.com/open-telemetry/opentelemetry-kotlin/issues/938), [#952](https://github.com/open-telemetry/opentelemetry-kotlin/issues/952), and the [View-related](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view) issues, then marked this issue as blocked by them.

Anyways, the ultimate goal is to implement concrete instruments. Before that, an instrument needs to be provided with a WritableMetricStorage:

```kotlin
internal class DoubleUpDownCounterImpl(
    override val descriptor: InstrumentDescriptor,
    private val storage: WritableMetricStorage,
    private val sdkErrorHandler: SdkErrorHandler,
) : DoubleUpDownCounter, SdkInstrument
```

The first missing piece is therefore something like a MetricStorageRegistrar (naming is negotiable):
```kotlin
val storage = metricStorageRegistrar.registerSynchronous(
    instrument = descriptor,
    instrumentationScopeInfo = instrumentationScopeInfo,
)
```

The registrar would implement part of the SDK’s [measurement-processing flow](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#measurement-processing) by resolving and registering storage for each configured reader and metric stream.

Implementing it requires several supporting components:
- MetricStorageRegistry (not registrar), an internal implementation detail for storing registered MetricStorages
- [MetricReader](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#metricreader) and its per-reader "pipeline"
[...]

PR #695, opened 2026-07-21
metrics: add public Instrument API

inetand:
## Goal

Add the foundation of the Instrument API from the [OTel metrics spec](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument).
Partially addresses: https://github.com/open-telemetry/opentelemetry-kotlin/issues/378

### Instrument / SynchronousInstrument / AsynchronousInstrument separation

This mirrors the specification's synchronous/asynchronous taxonomy.
The OpenTelemetry specification defines different capabilities for synchronous and asynchronous instruments. Modeling these as separate interfaces allows those capabilities to be expressed in the type system without exposing irrelevant operations. `isEnabled()` is specified for synchronous instruments, while the callback lifecycle (`close()`, and later `registerCallback`) belongs only to asynchronous instruments.

  ### Why not `Instrument<T : Number>`

If `UInt` were supported for instruments in the future, it would not fit this design because `UInt` is not a subtype of `Number`.

  ### Why not `Instrument<T>` / `Histogram<T>` 

A generic `record(value: T) ` compiles to `record(Object) ` on the JVM, requiring unboxing on every call. While this overhead is negligible at low call volumes, it becomes measurable at scale.

Local measurement on Android (median ns/op):

  | Type | Generic | Specialized | Difference |
  |---|---|---|---|
  | Long | 9.18 ns | 1.59 ns | 5.8× slower |
  | Double | 9.60 ns | 2.12 ns | 4.5× slower |

On my local machine (desktop JVM), HotSpot inlined the unboxing well — ART on Android (androidMain) does not. 
For this reason, typed instruments expose specialized methods such as DoubleHistogram.record(value: Double, ...) and LongHistogram.record(value: Long, ...).

  ### No `kind` in the public API

An instrument's kind is represented by its type. For example, a DoubleHistogram is, by definition, a synchronous histogram over doubles.
This is consistent with other OpenTelemetry SDKs (such as Java and .NET), which also do not expose an explicit instrument kind in the public API. The implementation, however, can still use kind and value type internally for aggregation.

  ### Instrument declaration proposal

Each instrument kind is declared in its own dedicated package. For example, histograms will live under io.opentelemetry.kotlin.metrics.histogram.

  ## Follow-ups

  - `AsynchronousInstrument.registerCallback` with an unregistration handle
  - Typed histogram instruments and their `Meter` creation functions


inetand:
This is the approach I originally tried while exploring a generic Histogram<T> API. I ultimately dropped it for the reasons described in the PR description. Leaving it here for reference.

```kotlin
public sealed interface Histogram<T> : SynchronousInstrument {  
  
   @ThreadSafe  
    public fun record(  
        value: T,  
        context: Context? = null,  
        attributes: (AttributesMutator.() -> Unit)? = null,  
    )  
}  

// May be placed in different files
public interface LongHistogram : Histogram<Long>  
public interface DoubleHistogram : Histogram<Double>  
public interface UIntHistogram : Histogram<UInt>
```

api/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/SynchronousInstrument.kt
```diff
@@ -0,0 +1,23 @@
...
+public interface SynchronousInstrument : Instrument {
+
+    /**
+     * Returns whether this instrument is currently enabled.
+     *
+     * The result may change over time: instrumentation should call this each time a measurement
+     * is recorded, immediately before performing expensive work needed only to produce the
+     * measurement.
+     */
+    public fun isEnabled(): Boolean
```
inetand: I've realized that Logger uses enabled(), but I instinctively named this isEnabled(), following the style used by [Kotlin's Boolean APIs, such as isEmpty(), isBlank() etc. ](https://kotlinlang.org/docs/booleans.html#boolean-in-conditions)

Would it be better to align with Logger and rename it to enabled()?

fractalwrench: I'd lean towards `enabled` as that seems to be the OpenTelemetry convention.

api/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/Instrument.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Records measurements which are aggregated to metrics.
+ *
+ * Instruments are obtained from a [Meter] and are identified by their [name], kind, [unit],
+ * and [description].
+ *
+ * https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument
+ */
+@ExperimentalApi
+@ThreadSafe
+public interface Instrument {
```
fractalwrench, line 16: The advisory parameter seems to be missing. Is that intentional? https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-advisory-parameters

fractalwrench (approved):
LGTM with one question. Thanks for the thorough discussion around design choices; it helped. I'm happy with this as an initial starting point and it can be iterated upon if the implementations don't quite fit...

api/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/Instrument.kt
inetand: Oh, I completely forgot to mention Advice in the PR description, thanks for pointing that out!

This was intentional for two reasons:

1. Since the spec defines only name, unit, and description as an instrument's identifying properties, I decided to expose Advice only as a parameter when creating an instrument, not through the Instrument interface.
2. It can always be added to the interface later if there's demand for it, but it seemed better to keep the initial version of the API narrow.

I should have included this earlier, but here's the intended way to declare an instrument:

```kotlin
public interface Meter {

    public fun doubleHistogram(
        name: String,
        unit: String? = null,
        description: String? = null,
        bucketBoundariesAdvice: List<Double>? = null,
        attributeKeysAdvice: List<String>? = null,
    ): DoubleHistogram
}

api/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/SynchronousInstrument.kt
inetand: I see, that makes sense. I'll rename it to enabled() to align with the OTel convention

api/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/Instrument.kt
fractalwrench: Got it. That approach makes sense to me 👍

fractalwrench:
@inetand asking here as I wasn't sure if you were on the CNCF slack, but would you be interested in 'triager' status on opentelemetry-kotlin?

The role (and pathway to other roles) is explained here, and I'm happy to clarify any questions you might have: https://github.com/open-telemetry/community/blob/main/guides/contributor/membership.md#membership-levels

Thanks again for your contributions! If you're interested let me or Jason know, and we'll kick off the process.

inetand:
@fractalwrench Thanks for the invitation! I'd love to help the project evolve by contributing as a triager!

I just have one question: my availability may vary depending on work, so there may be periods when I'm not very active. Is that okay for an OpenTelemetry triager, or is there an expectation of regular activity?

P.S. Regarding Slack, I had already joined the CNCF Slack workspace, but I wasn't in the OpenTelemetry Kotlin channel yet. Thanks for reaching out here–I really appreciate it! I've joined the channel now as well.

PR #697, opened 2026-07-21
turn gradle tooling parallel on

breedx-splk:
New feature in gradle allows parallel gradle sync in IDEs without impacting the main production builds.

https://docs.gradle.org/current/userguide/performance.html#sec:configure_tooling_api_actions_parallelism

Seems like a reasonable thing to me.


fractalwrench (approved):
LGTM. We can monitor if any build problems do crop up and disable this flag if so.

PR #698, opened 2026-07-21
Use environment secrets for publishing

breedx-splk:
This allows us to only have the secrets defined for the `protected` environment, which is then only used for publishing jobs. This reduces the surface area of jobs for which the secrets apply and improves the security posture of other non-publishing jobs.


fractalwrench (approved):
LGTM

PR #696, opened 2026-07-21
feat(Tracing API): Implement Composite Sampler

jdvr:
## Goal

Closes #659

Implements the `CompositeSampler` from the [OTel spec](https://opentelemetry.io/docs/specs/otel/trace/sdk/#compositesampler), along with the built-in `ComposableSamplers`.

### `sdk-api` (common, all targets)
- New ComposableSampler interface: a sampler that returns a SamplingIntent instead of a final decision
- New SamplingIntent interface: carries threshold, adjusted-count reliability, optional attributes, and optional TraceState updater
- `composite {}` DSL entry point that wraps a ComposableSampler into a regular Sampler (see comments, the other option was `composite(...)`
- Built-in composable samplers: `composableAlwaysOn`, `composableAlwaysOff`, `composableProbability`, `composableParentThreshold`

### `compat` (JVM/Android)
- Bidirectional adapters bridging Kotlin ComposableSampler ↔ OTel Java ComposableSampler
- All four composable built-ins exposed on CompatBuiltInSamplerApi (delegating to OTel Java)
- New Java type aliases: OtelJavaComposableSampler, OtelJavaCompositeSampler, OtelJavaSamplingIntent

### Out of scope (follow-up)

`ComposableRuleBased` and `ComposableAnnotating` are not included to keep this PR reviewable. They can be added as a follow-up.

## Testing

- Unit tests covering all new code
- Example app updated to demonstrate the composite {} sampler DSL

Issue #659, opened 2026-07-14
Implement composite sampler


fractalwrench:
Implement a composite sampler as per the spec: https://opentelemetry.io/docs/specs/otel/trace/sdk/#compositesampler

See `PropagatorConfigDsl#composite()` for inspiration

jdvr:
I want to work on this 🙏 😄

fractalwrench:
Thanks @jdvr - I've assigned the issue to you.

PR #708, opened 2026-07-28
feat(Tracing API): Implement CompositeSampler and built-in composable samplers

jdvr:
## Goal

Implements the `CompositeSampler` and `ComposableSampler` abstractions from the OTel spec (https://opentelemetry.io/docs/specs/otel/trace/sdk/#compositesampler), along with the built-in composable samplers: ComposableAlwaysOn, ComposableAlwaysOff, ComposableProbability, and ComposableParentThreshold.

Part of #659

Split from #696 

### `sdk-api` / `implementation` (common, all targets)
- New ComposableSampler interface: a sampler that returns a SamplingIntent instead of a final decision
- New SamplingIntent interface: carries threshold, adjusted-count reliability, optional attributes, and optional TraceState updater
- `composite {}` DSL entry point that wraps a ComposableSampler into a regular Sampler (see comments, the other option was `composite(...)`
- Built-in composable samplers: `composableAlwaysOn`, `composableAlwaysOff`, `composableProbability`, `composableParentThreshold`

### Out of scope (follow-up)

- [ ] `ComposableRuleBased` and `ComposableAnnotating` are not included to keep this PR reviewable. They can be added as a follow-up.

- [x] `compat` and `examples` added as follow up PR #709 

## Testing

- Unit tests covering all new code

### DIFF
--- a/implementation/api/jvm/implementation.api
+++ b/implementation/api/jvm/implementation.api
@@ -7,7 +7,43 @@ public final class io/opentelemetry/kotlin/tracing/sampling/BuiltInSamplerApiKt 
 	public static final fun alwaysOff (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;)Lio/opentelemetry/kotlin/tracing/sampling/Sampler;
 	public static final fun alwaysOn (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;)Lio/opentelemetry/kotlin/tracing/sampling/Sampler;
 	public static final fun alwaysRecord (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;)Lio/opentelemetry/kotlin/tracing/sampling/Sampler;
+	public static final fun composableAlwaysOff (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;)Lio/opentelemetry/kotlin/tracing/sampling/ComposableSampler;
+	public static final fun composableAlwaysOn (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;)Lio/opentelemetry/kotlin/tracing/sampling/ComposableSampler;
+	public static final fun composableParentThreshold (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;Lio/opentelemetry/kotlin/tracing/sampling/ComposableSampler;)Lio/opentelemetry/kotlin/tracing/sampling/ComposableSampler;
+	public static final fun composableProbability (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;D)Lio/opentelemetry/kotlin/tracing/sampling/ComposableSampler;
+	public static final fun composite (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;Lkotlin/jvm/functions/Function1;)Lio/opentelemetry/kotlin/tracing/sampling/Sampler;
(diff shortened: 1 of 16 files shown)


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/BuiltInSamplerApi.kt
```diff
@@ -47,3 +47,52 @@ public fun SamplerConfigDsl.parentBased(
...
+/**
+ * Configures sampling by delegating to a [ComposableSampler], using consistent probability
+ * sampling over the OpenTelemetry TraceState `ot` `th`/`rv` sub-keys.
+ *
+ * https://opentelemetry.io/docs/specs/otel/trace/sdk/#compositesampler
+ */
+@ExperimentalApi
+public fun SamplerConfigDsl.composite(block: SamplerConfigDsl.() -> ComposableSampler): Sampler =
+    CompositeSampler(block())
+
```
jdvr, line 60: This allows to do `composite {` instead of `composite(`, I think this feels more idiomatic specifically when we add support for ruled-based sampler:

```kotlin
sampler {
   composite {
     alwaysOff()
   }
}
// vs

sampler {
   composite(alwaysOff())
}
```


This will also have more fluet API for rule-based sampler.

fractalwrench: Agree that this syntax looks better than the alternative

implementation/api/jvm/implementation.api
```diff
@@ -7,7 +7,43 @@ public final class io/opentelemetry/kotlin/tracing/sampling/BuiltInSamplerApiKt
...
+	public static final fun composite (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;Lkotlin/jvm/functions/Function1;)Lio/opentelemetry/kotlin/tracing/sampling/Sampler;
 	public static final fun parentBased (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;)Lio/opentelemetry/kotlin/tracing/sampling/Sampler;
 	public static synthetic fun parentBased$default (Lio/opentelemetry/kotlin/init/SamplerConfigDsl;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;Lio/opentelemetry/kotlin/tracing/sampling/Sampler;ILjava/lang/Object;)Lio/opentelemetry/kotlin/tracing/sampling/Sampler;
 }
 
+public final class io/opentelemetry/kotlin/tracing/sampling/ComposableAlwaysOffSampler : io/opentelemetry/kotlin/tracing/sampling/ComposableSampler {
+	public fun <init> ()V
```
fractalwrench: Is there a way this could be defined as a function or interface instead? Or perhaps have its visibility set to internal? One of the guidelines in `CONTRIBUTING.md` is to avoid concrete types in the public API, as it allows implementation details to leak: https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/CONTRIBUTING.md#development-guidelines

fractalwrench (commented):
Thanks! One question around the visibility of some classes, but otherwise this looks good from my POV.

implementation/api/jvm/implementation.api
jdvr: You are right, I have made all this internal 🙇

fractalwrench (approved):
LGTM - thanks!

PR #710, opened 2026-07-28
Allow configuring SDK error handler

### LINKED ISSUE #671: Provide mechanism for setting custom error handler
The spec states that a mechanism for setting a custom error handler should be provided: https://opentelemetry.io/docs/specs/otel/error-handling/#configuring-error-handlers

IMO this should be done on the `OpenTelemetryConfigDsl` (using similar syntax to the approach of `ImplicitContextStorage`). The `SdkErrorHandler` interface should be reviewed as part of this and potentially altered/bridged.

fractalwrench:
## Goal

Allows library consumers to configure a custom SDK error handler. Closes #671.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. Makes sense to genericize the interface and make the SDK implementation sort the different errors out itself.

PR #716, opened 2026-07-30
Correctly link child span to parent span when Span is decorated

fractalwrench:
## Goal

If `Span` is decorated in the `compat` implementation the child span doesn't correctly reference its parent span when `Span.storeInContext` is called. This is because that function does an explicit cast to `SpanAdapter` rather than `toOtelJavaContext()`.

I've resolved this problem and tidied up a few additional cases when `OtelJavaContextAdapter` was being constructed directly, rather than using `toOtelJavaContext()`.


lemnik (approved):
LGTM

bidetofevil (approved):
LGTM

PR #718, opened 2026-07-30
build: prevent r8 warning

fractalwrench:
## Goal

`opentelemetry-kotlin` 0.6.0 adds a new runtime dependency on `io.opentelemetry:opentelemetry-extension-trace-propagators`, which has classes that implement `io.opentelemetry.sdk.autoconfigure.spi.ConfigurablePropagatorProvider`.  This causes R8 warnings when compiling, such as [this one](https://github.com/embrace-io/embrace-android-sdk/actions/runs/30540000440/job/90865197591).

I've added a rule to the R8 consumer proguard files so that library users don't run into this.

Originally spotted here: https://github.com/embrace-io/embrace-android-sdk/pull/3702


bidetofevil (approved):
LGTM

Issue #652, opened 2026-07-14
Stabilize Logging API


fractalwrench:
Discuss what needs to change on the Logging API (if anything) before we remove the `ExperimentalApi` annotation.

PR #747, opened 2026-08-04
Update logger Kdoc

fractalwrench:
## Goal

Update logger Kdoc to clarify that `enabled()` should not have its return value cached. I've used the same wording as we did for `Tracer#enabled()`

Related to #652


breedx-splk (approved):
Cool, thanks for coming back to it. 👍🏻

PR #753, opened 2026-08-05
Add section to community roles

fractalwrench:
## Goal

Adds a section to the community roles explaining that maintainers will periodicially review roles. My intent here was to give a bit more transparency in how the roles are decided, and also to add a note that will encourage folks to get in touch if they're unsure on whether they're meeting a role's requirements or not.


breedx-splk (approved):
Seems generous but I like it. :)

PR #754, opened 2026-08-06
Set a schema url on the default resource.

### LINKED ISSUE #736: Set schema URL on default resource
The schema URL on the default resource created by the SDK in `ResourceConfigImpl#sdkDefaultResource()` should be set [according to the spec](https://opentelemetry.io/docs/specs/otel/resource/sdk/#detecting-resource-information-from-the-environment). This could be achieved by using `BuildKonfig` to create a build-time constant of the semconv version.

fractalwrench:
## Goal

Sets a schema URL on the SDK's default resource. Closes #736.

## Testing

Added unit tests.


priettt (approved):
LGTM! 

SEMCONV_VERSION seems unused, maybe it's used on a following PR?

fractalwrench:
SEMCONV_VERSION is unused - it's from an earlier iteration. I'll remove it.

PR #755, opened 2026-08-07
Let a custom IdGenerator declare that it produces random trace IDs

### LINKED ISSUE #552: Set random flag on TraceFlags correctly

### LINKED ISSUE #724: Let a custom `IdGenerator` declare that it produces random trace IDs
IdGenerator should declare that it produces random trace IDs via a boolean flag on the interface. This can be used by the SDK so set the `random` flag appropriately.
https://opentelemetry.io/docs/specs/otel/trace/sdk/#idgenerator-randomness

fractalwrench:
## Goal

Closes #724 by allowing `IdGenerator` implementations to specify whether they support W3C randomness. Additionally closes #552.

## Testing

Added unit tests


priettt (approved):
Looks good 🙌

PR #756, opened 2026-08-07
Clarify schemaUrl behavior if merged

### LINKED ISSUE #737: Resource merge silently resolves conflicting schema URLs
`Resource.merge` silently resolves conflicting schema URLs by picking `other.schemaUrl`. We should report an `SdkError` if two resource objects have different schemaUrls and note the merge behavior in the Kdoc.

fractalwrench:
## Goal

Closes #737 by documenting the behavior for if two resource objects have a schemaUrl mismatch. The behavior I've chosen is to avoid specifying a schemaUrl, as otherwise the SDK is being misleading.

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/OpenTelemetryImplEntrypoint.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/OpenTelemetryImplEntrypoint.kt
@@ -32,7 +32,7 @@ public fun createOpenTelemetry(
     val cfg = OpenTelemetryConfigImpl(clock).apply(config)
     val idGenerator = cfg.resolveIdGenerator()
 
-    val resourceFactory = ResourceFactoryImpl()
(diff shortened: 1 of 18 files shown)


priettt (commented):
@fractalwrench Wouldn't we always log a warning due to #754?

fractalwrench:
@priettt good question. Currently the spec states that if two resources have non-empty schema URLs and a merge is attempted, the behavior is undefined: https://opentelemetry.io/docs/specs/otel/resource/sdk/#merge-behavior-without-entities

After re-reading the spec, I'm inclined to believe that the current behavior (where the schema URL is overwritten) feels more correct, particularly when considering the default resource sets its own schema URL. What are your thoughts?

priettt:
@fractalwrench I agree! If this ever becomes an issue we can revisit in the future, or maybe add some logging, but I doubt it will.

bidetofevil (approved):
LGTM

fractalwrench:
Closing out in favour of current behavior for now.

PR #757, opened 2026-08-07
Fix conversion of AnyValue

### LINKED ISSUE #723: attrsFromMap handles AnyValue incorrectly
`attrsFromMap` handles `AnyValue` incorrectly. Its values are currently serialized as `LongValue(value=3)` rather than `3`:
https://github.com/open-telemetry/opentelemetry-kotlin/blob/53703f66033cec9922341b5fbb9a2434278df511/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/attributes/OtelJavaAttributesExt.kt#L11

We should fix that and also check whether anywhere else has the same problem.

fractalwrench:
## Goal

Closes #723 by fixing how `AnyValue` is converted when using the compat layer.


priettt (approved):
👍

PR #763, opened 2026-08-07
Guard SdkErrorHandler

fractalwrench:
## Goal

Adds a guard around `SdkErrorHandler` so that when implementations throw, the SDK won't terminate the process.


priettt (approved):
👍

Issue #762, opened 2026-08-07
Create intermediate model layer for configuration


fractalwrench:
We need an intermediate layer for configuration that acts as the source of truth. Currently the SDK has the following layers:

1. Config on DSL objects (tied to live models that affect tracer/logger config on edits)
2. Config in envars (separate classes, mostly unimplemented)
3. Config in yaml (relies on config-schema, unimplemented)

The DSL should take precedence over yaml or envars. Additionally according to the spec, envars should be ignored if yaml is present.

This implementation would be more straightforward if we had an intermediate config model that simply holds data on how the SDK should be configured. Holding config in this way would allow:

1. The DSL to avoid altering SDK objects until the SDK is initialized, saving potentially expensive operations
2. YAML and envar to use the same config model, making any precedence/merge rules straightforward to implement

PR #764, opened 2026-08-07
Create config model layer

fractalwrench:
## Goal

Begins working towards #762 by adding a `config-model` module with a model class representation of how OpenTelemetry can be configured. This is independent of the DSL (because that acts on live objects) and envar/yaml parsing, so that the concerns are separated.

Separating out this logic should make it simpler to apply precedence rules for config and also to reason about how the SDK is behaving.

## Testing

Added unit tests.

### DIFF
--- a/config-model/README.md
+++ b/config-model/README.md
@@ -1,0 +1,15 @@
+# config-model
+
+This module contains the intermediate configuration model that every configuration mechanism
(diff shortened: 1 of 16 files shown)


priettt (approved):
Nice 🙌

PR #770, opened 2026-08-10
Guard call to enabled() in processors

fractalwrench:
## Goal

Ensures that `enabled()` is guarded in user-submitted processors so that if it throws execution still continues. I've chosen to assume that throwing means the processor should continue as normal.

### DIFF
--- a/implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/BareProcessorFaultIsolationTest.kt
+++ b/implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/BareProcessorFaultIsolationTest.kt
@@ -1,0 +1,113 @@
+package io.opentelemetry.kotlin
+
+import io.opentelemetry.kotlin.error.FakeSdkErrorHandler
+import io.opentelemetry.kotlin.error.SdkErrorSeverity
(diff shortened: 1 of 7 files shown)


implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/BareProcessorFaultIsolationTest.kt
```diff
@@ -0,0 +1,113 @@
...
+    fun `a bare log record processor throwing in enabled does not reach the caller`() {
+        val processor = FakeLogRecordProcessor(enabledResult = { error("boom") })
+        val otel = createOpenTelemetry {
+            errorHandler(handler)
+            loggerProvider {
+                export { processor }
+            }
+        }
+
+        assertFalse(otel.loggerProvider.getLogger("test").enabled())
```
priettt, line 97: Doesn't this conflict with "throwing means the processor should continue as normal"?

LogRecordProcessor.enabled() defaults to true, so if a processor throws, shouldn't enable return true as if nothing happened?

fractalwrench:
Closing in favour of #771

PR #771, opened 2026-08-10
Add guards around processors

fractalwrench:
## Goal

Adds a try-catch around processors as these are user-submitted code that could potentially throw. The SDK should be resilient and continue sending telemetry in this instance.

## Testing

Added unit tests.


priettt:
CI is failing.

Also, doesn't this conflict with the #770 implementation? Should we pick this one over the other?

bidetofevil (approved):
Very nice. LGTM

PR #774, opened 2026-08-10
Read envars on non-jvm platforms

### LINKED ISSUE #443: Retrieve environment variables on non-JVM platforms
## Goal

Currently environment variables are only read on JVM platforms via `System.getEnv()`. An implementation should be added (eventually) for non-JVM platforms that are likely to use this, such as nodejs.

fractalwrench:
## Goal

Closes #443 by reading environment variables correctly on non-JVM platforms.

## Testing

Added tests.


bidetofevil (approved):
LGTM

Issue #735, opened 2026-08-03
Create ComposableRuleBased and ComposableAnnotating


fractalwrench:
Create [ComposableRuleBased](https://opentelemetry.io/docs/specs/otel/trace/sdk/#composablerulebased) and [ComposableAnnotating](https://opentelemetry.io/docs/specs/otel/trace/sdk/#composableannotating) as per the spec.

PR #775, opened 2026-08-10
Implement composable rules sampler

fractalwrench:
## Goal

Adds a [composable rules-based sampler](https://opentelemetry.io/docs/specs/otel/trace/sdk/#composablerulebased), partially addressing #735.

This allows complex rules to be built for whether spans should be sampled or not. For example, the following would specify that spans named '/health' should be dropped, but spans with 'SERVER' SpanKind should always be sampled:

```kotlin
val sampler = samplerDsl.composableRuleBased {
    rule({ _, name, _, _, _ -> name == "/health" }) { composableAlwaysOff() }
    rule({ _, _, spanKind, _, _ -> spanKind == SpanKind.SERVER }) { composableAlwaysOn() }
}
```

## Testing

Added unit tests.


sdk-api/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/SamplingPredicate.kt
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * https://opentelemetry.io/docs/specs/otel/trace/sdk/#composablerulebased
+ */
+@ExperimentalApi
+public fun interface SamplingPredicate {
+
+    /**
+     * Returns true if the rule guarded by this predicate should decide how the span is sampled.
+     */
+    public fun matches(
```
bidetofevil, line 23: Any reason span events aren't included in the input?

bidetofevil (approved):
LGTM. One comment about span events. Non-blocking.

sdk-api/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/SamplingPredicate.kt
fractalwrench: Spans don't have events at the time of sampling

Issue #66, opened 2026-01-20
Create HTTP request instrumentation for Ktor


fractalwrench:
## Goal

The project does not have any optional instrumentation written as of yet. Ktor is a good example as it's widely used in KMP projects and the semantic conventions for HTTP requests is already stable.

HylkeB:
I'm working on my own ktor instrumentation right now (focussed only on the mobile client). Any progress on this? 

FYI, my approach is creating a custom ktor plugin (hooked to the send hook) that extracts request and response data (or exception) from the call. Then the tricky part is properly extracting the values required according to the semantic conventions of open telemetry.

Things I did:
- Ensure traceparent is included in the request headers for [context propagation](https://opentelemetry.io/docs/concepts/context-propagation/) 
- Read traceparent header to create the spancontext
- extract the [spanName](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name) in the form of "{method} {low-cardinality path}" 
  - I determined low cardinality path by replacing any segment that looks like an identifier into *** (currently only number path segments and uuids). So `https://example.com/orders/103478/items/2934` becomes `/orders/***/items/***`
- extract [arguments](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span) from the request and response data according the semantic conventions
  - as far as possible, i tried to apply the required / conditionally required fields, and included some recommended and opt-in fields as well, as long as they were stable.
- extract [status](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status) from the result
  - as per instructions, never going to OK and leaving the error description empty as its already in the arguments error.type

I also wonder, how are you planning on extracting the `url.template` from ktor to properly fill the span name? I guess that largely depends on conventions of the app developers or maybe some official libraries such as retrofit provide the value for template in ktor request attributes?

fractalwrench:
@HylkeB there was an attempt in #268 but that seems to have stalled out. I'm not aware of anyone else who is actively working on it. If you're able, donating Ktor instrumentation would be greatly appreciated.

PR #777, opened 2026-08-10
Start building instrumentation for Ktor

fractalwrench:
## Goal

Partially addresses #66 by starting some instrumentation for [Ktor](https://ktor.io/). Currently this just creates a very simple trace for a HTTP request on both the client and the server, and adds all the boilerplate module to setup the instrumentation.

Why instrumentation? Ktor feels like a good first candidate as it's a widely used KMP library that supports all our targets, and there are already [well established semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/) for HTTP requests. I also think the project is at a point where it would be helpful to have examples of how the API can be used to write instrumentation.

I've kept this changeset to the minimum possible, but in future I envision the following improvements:

- Capture all other relevant attributes for HTTP requests
- Inject a traceparent header
- Setup the example app to use Ktor

## Testing

Added unit tests


Msksgm (approved):
LGTM

PR #778, opened 2026-08-10
Test the entire published surface at min supported versions and hold deps at the Kotlin 2.0 floor

priettt:
## Goal

The min supported versions fixtures consumed a fixed list of five modules, so modules added later were never checked against the declared floors.

The build now derives the modules automatically and the fixtures consume every published module, which means a new module joins the check automatically. 

Covering the full surface revealed two breaches of the `minSupportedKotlinLang = 2.0` floor. A Kotlin compiler can read metadata at most one minor version ahead of itself, and:

- `config-schema` exposes kotlinx-serialization 1.11.0 as an `api` dependency. Its classes and its kotlin-stdlib 2.3.20 requirement both carry metadata 2.3.0, which compilers below Kotlin 2.2 cannot read.
- `exporters-otlp` uses ktor 3.5.2, which pulls kotlin-stdlib 2.3.21 into consumer classpaths. On Android this reaches the compile classpath as well, because AGP aligns it with the runtime classpath.

The second commit pins both at the newest versions compatible with the floor: kotlinx-serialization 1.8.1 (kotlin-stdlib 2.1.20) and ktor 3.2.3 (kotlin-stdlib 2.1.21, and it requires exactly kotlinx-serialization-json 1.8.1).


fractalwrench (approved):
LGTM - thanks!

PR #780, opened 2026-08-11
Reorganise config into separate modules

fractalwrench:
## Goal

Organises the configuration code into separate modules:

`config` - entrypoint for obtained a resolved configuration object. The implementation and compat module both call this, then apply limits/behavior using its objects
`config-dsl` - implementations of the config DSL that hold values. Decouples the DSL so it doesn't cause side effects until the SDK initializes
`config-envars` - reads config from environment variables
`config-yaml` - reads config from a declarative YAML file
`behaviorl` - an intermediate model layer that maps envars, yaml, and DSL config onto the same type, and then applies precedence rules
`config-schema` - generated classes from [opentelemetry-configuration](https://github.com/open-telemetry/opentelemetry-configuration) used to deserialize yaml

This changeset doesn't have any functional changes and simply reorganises code. A future changeset will build on top of these changes and demonstrate how config should be implemented, using the span limits as an example.


bidetofevil (approved):
Very nice abstraction. LGTM

PR #803, opened 2026-08-11
fix: remove by lazy on SpanModel and LogRecordModel lock

### LINKED ISSUE #797: Remove by lazy on SpanModel lock
The `by lazy` on SpanModel is redundant as the lock is almost always touched on the first property access, and removing it reduces object allocations & a volatile access in lazy.

Msksgm:
## Goal

Close #797.

Every read and mutation on `SpanModel` goes through `lock.read {}` / `lock.write {}`, so `by lazy` on the lock never actually defers anything — it just costs an extra `SynchronizedLazyImpl` per span and a volatile read per access. `LogRecordModel` had the same pattern, so it is fixed too. Both classes are `internal`, so no API change.

## Testing

JMH GC profiler on `benchmark-jvm`: `SimpleLoggingBenchmark` drops from 112 to 88 B/op — exactly one `SynchronizedLazyImpl`. `CreateSimpleSpanBenchmark` drops by the same 24 B/op. Throughput is unchanged within error bars.

`jvmTest` across all modules and detekt on the changed module pass.

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/model/LogRecordModel.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/model/LogRecordModel.kt
@@ -26,9 +26,7 @@ internal class LogRecordModel(
     logLimitConfig: LogLimitConfig,
 ) : ReadWriteLogRecord {
 
-    private val lock by lazy {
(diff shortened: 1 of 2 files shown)


fractalwrench (approved):
LGTM - thanks!

PR #811, opened 2026-08-12
Optimize allocations for samplers

### LINKED ISSUE #795: Make AlwaysOnSampler and AlwaysOffSampler objects
AlwaysOnSampler and AlwaysOffSampler can be objects to avoid unnecessary allocations

### LINKED ISSUE #807: Samplers should use empty attribute container
Samplers that don't set attributes (most of the built-ins) should avoid allocating a new AttributesModel for every sampling decision

fractalwrench:
## Goal

The built-in samplers were constructing a new AttributeContainer every time, when most of them just supply an empty map. Additionally, always on/off samplers can be implemented using just one object, rather than creating a new instance for each piece of telemetry.

Closes #807
Closes #795


Msksgm (approved):
LGTM

bidetofevil (approved):
LGTM

PR #812, opened 2026-08-12
Fix potential key collision in providers

### LINKED ISSUE #802: TracerProviderAdapter cache key collision
TracerProviderAdapter concats properties without a delimiter so `getTracer("ab", "c")` and `getTracer("a", "bc")` could collide.

fractalwrench:
## Goal

Closes #802 where the `TracerProvider` (and other providers) in compat could theoretically collide keys.


bidetofevil (approved):
LGTM

PR #813, opened 2026-08-12
Rethrow from wrapOperation

### LINKED ISSUE #793: wrapOperation should rethrow the exception
`wrapOperation` in `SpanExt` should rethrow the `Throwable` after recording it so that user code can react accordingly.

fractalwrench:
## Goal

Fixes #793 - wrapOperation didn't rethrow the exception it caught, which unexpectedly swallows throwables from user-supplied code.


bidetofevil (approved):
LGTM

PR #814, opened 2026-08-12
Optimize attribute truncation

### LINKED ISSUE #806: Attribute truncation fast path
The default `attributeValueLengthLimit` is `Int.MAX_VALUE`, a fast-path in
`truncateAnyValue` and `truncateString` would avoid unnecessary work & allocations.

fractalwrench:
## Goal

Closes #806 by optimizing the truncation of attributes so that when no limit is set, no new string instance needs to be constructed.


bidetofevil (approved):
LGTM

PR #815, opened 2026-08-12
Lock on span model properties

### LINKED ISSUE #801: SpanModel state property should be volatile or atomic
`state` is a non-`@Volatile` plain `var` read outside the lock by `isRecording()` and `hasEnded`. `endTimestamp` is also affected.

fractalwrench:
## Goal

Closes #801 by ensuring that when checking whether a span is recording, or accessing its timestamp, the lock is acquired so the state is guaranteed to be consistent. Additionally, `toSpanData()` is synchronized behind a single lock rather than multiple locks taken on individual accessors.


bidetofevil (approved):
Nice. LGTM

Issue #787, opened 2026-08-11
Live mutable models are retained rather than immutable snapshots


fractalwrench:
Processors currently don't call `toSpanData()` and reuse the `SpanModel` object with a view. It would be preferable to instantiate a new `SpanDataImpl` object during `onEnd` as this releases unused references within the `SpanModel` (such as locks, properties, etc).

The same change should be made to log records too.

PR #816, opened 2026-08-12
Take immutable snapshot of span when ended

fractalwrench:
## Goal

Partially addresses #787 by taking an immutable snapshot of the span for the processor. This releases all references to the span model allowing GC of the refs.

## Testing


bidetofevil (approved):
Familiar! LGTM

PR #817, opened 2026-08-12
Supply traceId as bytes to Sampler

### LINKED ISSUE #794: Improve trace ID hex encoding
IdGenerator allocates a string per byte, resulting in unnecessary allocations & garbage:

```kotlin
public fun ByteArray.toHexString(): String {
    val result = StringBuilder(size * 2)
    for (b in this) {
        val i = b.toInt() and 0xFF
        result.append(i.toString(16).padStart(2, '0'))
    }
    return result.toString()
}
```

We could attempt the following changes:

- Consider altering `Sampler.shouldSample` to take `traceIdBytes: ByteArray` and derive
  randomness from bytes (would need to check the spec)
- Rewrite `toHexString` with a lookup table into a `CharArray`
- Pass computed hex into `SpanContextImpl` instead of re-deriving

fractalwrench:
## Goal

`Sampler` in the spec doesn't specify [how `traceId` should be supplied](https://opentelemetry.io/docs/specs/otel/trace/sdk/#sampling), so I've switched the type from `String` to `ByteArray` as I believe this better fits the intent, and avoids needing to encode then decode a hex string in the implementation module.

I've additionally optimized the hex encoding so that fewer strings are allocated for each traceId/spanId hex string.

Closes #794

## Testing

Added unit tests & relied on existing coverage.


bidetofevil (approved):
As long as the spec conscious folks are ok with this, I'm good.

PR #818, opened 2026-08-12
Make scalars on config DSL nullable

fractalwrench:
## Goal

Makes the scalars on the config DSL nullable, so that they don't override default values in the SDK if already set. This is a pre-requisite refactor that will help merging config values set on the DSL with yaml/envars, which will land in future PRs.

## Testing

Updated existing tests.


bidetofevil (approved):
LGTM

PR #819, opened 2026-08-12
Sanitize SpanContext

### LINKED ISSUE #790: calculateSpanContext always returns valid
`calculateSpanContext` always returns valid as true. It should check the traceId when deriving this value.

fractalwrench:
## Goal

Closes #790 by deriving `isValid` on `SpanContext` from the `traceId` property, rather than always following what was supplied. Additionally adds some edge-case handling where invalid IDs might be supplied by a custom `IdGenerator`.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #822, opened 2026-08-13
Make immutable copy when exporting log record

### LINKED ISSUE #787: Live mutable models are retained rather than immutable snapshots

fractalwrench:
## Goal

Takes a similar approach to #816 by taking an immutable copy of the `LogRecord` when sending it to an exporter. This releases unused references on the `LogRecordModel`.

Closes #787


bidetofevil (approved):
LGTM

Issue #786, opened 2026-08-11
Protobuf conversion throws on edge-cases


fractalwrench:
1. A trace_state entry in `parseTraceState` in `CommonProtobufConversion` without `=` throws IndexOutOfBoundsException
2. `convertAttributeValue` in `AttributesProtobufConversion` throws UnsupportedOperationException for any type outside the allowlist. Stringify instead.
3. `DeserializedResource.asNewResource` and `DeserializedTraceState.put`/`remove` throw `UnsupportedOperationException`

PR #825, opened 2026-08-13
Address IndexOutOfBoundsException in protobuf deserialization

fractalwrench:
## Goal

Partially addresses #786 by catching an `IndexOutOfBoundsException` if a trace state entry doesn't have an `=` sign. The fix involved consolidation as trace state was handled differently in two different parts of the SDK.


bidetofevil (approved):
LGTM

PR #826, opened 2026-08-13
Implement attribute limits behavior

fractalwrench:
## Goal

Implements a behavior class for attribute limits. This is not currently hooked up to the SDK but will be used to drive its behavior with resolved config.


bidetofevil (approved):
LGTM

Issue #809, opened 2026-08-12
Improve BatchTelemetryProcessor implementation


fractalwrench:
`BatchTelemetryProcessor` could use several improvements:

~1. An ArrayDeque would reduce element moves when draining the queue compared to List~
2. The collection is mutated without a mutex
~3. The queue size check is off by one~
4. Export is not triggered when the maxExportBatchSize is exceeded
5. `forceFlush` returns success unconditionally
6. `shutdown` does not flush

See `BatchTelemetryLogRecordProcessor` for inspiration.

Rayan-and-beyond:
i took another pass through the repo guidance and their agent policy requires ai-generated code to be manually reviewed before committing. that doesn't fit this unattended worker lane, so i'm stepping back from this one rather than create work that can't follow the repo's contribution rules.

PR #827, opened 2026-08-13
Improve batch processor

fractalwrench:
## Goal

Uses `ArrayDeque` instead of `List` in batch processor to reduce the number of element moves, and fixes an off-by-one error. Partially addresses #809.


bidetofevil (approved):
LGTM

PR #828, opened 2026-08-13
Handle unknown attribute types in protobuf conversion

fractalwrench:
## Goal

Partially addresses #786 by handling unknown values when converting attribute values by stringifying, rather than throwing.


bidetofevil (approved):
LGTM

PR #829, opened 2026-08-13
Fix NPE in ContextAdapter

### LINKED ISSUE #782: ContextAdapter could throw NPE
`ContextAdapter` could throw NPEs when setting values on the context, as the type is T? but `value` is cast to a non-null type

fractalwrench:
## Goal

Closes #782 by fixing a potential NPE in `ContextAdapter`.


bidetofevil (approved):
LGTM

Issue #791, opened 2026-08-11
Starting/ending telemetry is unguarded


fractalwrench:
We should add a try-catch when starting/ending telemetry as a final fail-safe to avoid taking down the system if something goes wrong in creating the telemetry.

fractalwrench:
Partially addressed in #830, which covers the Logger/Tracer path. We still need:

1. More granular coverage at individual call-sites within Logger/Tracer, so that a single failure degrades the span, rather than taking it down entirely
2. Guards at the LogRecordModel/SpanModel layer
3. Guards in the in-built processors/exporters
4. Guards in the samplers

PR #830, opened 2026-08-13
Begin hardening SDK against uncaught errors

fractalwrench:
## Goal

Begins hardening the SDK against uncaught errors, partially addressing #791.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #831, opened 2026-08-13
Implement resource behavior

fractalwrench:
## Goal

Implements a behavior class for the resource. This is not currently hooked up to the SDK but will be used to drive its behavior with resolved config.


behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/ResourceBehavior.kt
```diff
@@ -0,0 +1,35 @@
+package io.opentelemetry.kotlin.behavior
+
+import io.opentelemetry.kotlin.ExperimentalApi
+
+/**
+ * The entity producing telemetry.
```
bidetofevil, line 6: Entity is probably a bad word choice here as this is not an OTel Entity 😅

behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/ResourceBehavior.kt
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * https://opentelemetry.io/docs/specs/otel/resource/sdk/
+ */
+@ExperimentalApi
+data class ResourceBehavior(
```
bidetofevil, line 11: You're trying to capture the "thing" that is using this SDK instance to record telemetry? Is this effectively the app if we're talking about Android? Why not just call it `Service` then to stick to the convention, even if it's a bit wrong for us?

behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/ResourceBehavior.kt
carlosalberto: This is actually wording from the Specification itself that predates Entities :D but yes, better move from that to be on the safe spot.

fractalwrench: I've tweaked the wording here to use 'service' instead

behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/ResourceBehavior.kt
fractalwrench: `ResourceBehavior` is somewhat analogous to how embrace-android-sdk processes configuration. It's meant to describe how the SDK should behave around resources and abstracts away the precedence rules for whether that was set via the DSL, envars, or YAML.

bidetofevil (approved):
LGTM.

PR #832, opened 2026-08-13
Wait for operation result code

fractalwrench:
## Goal

Waits for the operation result code. The current approach doesn't account for when `CompleteableResultCode` is an async operation, so we need to wait for a result in a coroutine.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #834, opened 2026-08-13
fix: Default AttributesModel to mutableMapOf

### LINKED ISSUE #810: Default AttributesModel to mutableMapOf
We should default `AttributesModel` to `mutableMapOf` rather than `ConcurrentHashMap` as this takes less in allocations and the concurrency frequently isn't required

Msksgm:
## Goal

Close https://github.com/open-telemetry/opentelemetry-kotlin/issues/810.

`AttributesModel` should not default to the platform-specific `threadSafeMap()`. Most instances are short-lived or are protected by their owning model, so the default backing map can be a plain `mutableMapOf()` without implying that `AttributesModel` itself is thread-safe.

This removes the `@ThreadSafe` marker from the internal implementation and leaves synchronization decisions to the owner/call site. Attribute behavior is otherwise unchanged: validation, overwrite behavior, truncation, limits, and dropped-attribute counting still use the same `AttributesModel` logic.

## Testing

- Not run yet

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesModel.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesModel.kt
@@ -1,13 +1,9 @@
 package io.opentelemetry.kotlin.attributes
 
-import io.opentelemetry.kotlin.ThreadSafe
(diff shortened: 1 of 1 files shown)


fractalwrench (approved):
LGTM - thanks!

PR #836, opened 2026-08-14
Read config from all sources for log limit

fractalwrench:
## Goal

Reads config from all sources (DSL, envars, yaml) for log limits. This is not hooked up to control the SDK yet, but now provides a good model of how config is intended to work. The envars/yaml are read and converted to `LogLimitsBehavior`, as is the DSL.

`LogLimitsBehavior` is then resolved according to its precedence rules and can be used to control the behavior of either implementations of the SDK.

## Testing

Added unit tests.


lemnik (approved):

bidetofevil (approved):
LGTM

PR #837, opened 2026-08-14
Remove context key API extension

### LINKED ISSUE #799: No way to read values back from ContextFactoryApiExt
In `ContextFactoryApiExt` `createKey(key)` is invoked, but as `ContextKeyImpl` has no `equals`/`hashCode` there's no way to ever read the value back.

fractalwrench:
## Goal

Closes #799 by removing a function in the api-ext module. This sets value on the `Context` but because `ContextKey` is never returned, it's impossible to ever read them back, making them useless.

I've therefore removed them from the API and updated the docs to state what the preferred approach is (inline with the OTel spec)


lemnik (approved):

bidetofevil (approved):
LGTM.

PR #838, opened 2026-08-14
Design API for configuring via yaml

fractalwrench:
## Goal

Designs an API on the DSL for supplying a YAML config file.

### DIFF
--- a/sdk-api/src/commonMain/kotlin/io/opentelemetry/kotlin/init/ConfigFileDsl.kt
+++ b/sdk-api/src/commonMain/kotlin/io/opentelemetry/kotlin/init/ConfigFileDsl.kt
@@ -1,0 +1,21 @@
+package io.opentelemetry.kotlin.init
+
+import io.opentelemetry.kotlin.ExperimentalApi
+
+/**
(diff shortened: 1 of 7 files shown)


lemnik (approved):

sdk-api/src/commonMain/kotlin/io/opentelemetry/kotlin/init/ConfigFileDsl.kt
```diff
@@ -0,0 +1,19 @@
...
+@ExperimentalApi
+public interface ConfigFileDsl {
+
+    /**
+     * Declares the path of a YAML configuration file, in the format defined by
+     * opentelemetry-configuration. Declaring more than one file retains the last one declared.
+     *
+     * The SDK does not read or act on the file yet, so this API has no effect.
+     */
+    public fun configFile(path: String)
```
bidetofevil, line 20: Do we need to say anything about what kind of path we will resolve? i.e. both relative and absolute or just one of them?

bidetofevil (approved):
LGTM. Non-blocking comment

Issue #789, opened 2026-08-11
Fresh HttpClient created per exporter


fractalwrench:
Every `otlpHttpSpanExporter(...)` / `otlpHttpLogRecordExporter(...)` currently constructs its own
`HttpClient` + engine. One client should be shared across all by default.

PR #843, opened 2026-08-16
Add exporter fix to prevent HttpClient re-creation - reusable registry for engines+timeout

RamachandraBhardwaj:
## Goal

Fix the duplicate `HttpClient` Creation in exporters(`OtlpClient` specifically) - using registry pattern.

### Problem :

Each call to otlpHttpSpanExporter(...) / otlpHttpLogRecordExporter(...) could create a new HttpClient + engine path, increasing resource usage and causing avoidable client duplication.

### Solution:
Introduce a shared registry keyed by (engine, timeoutMs) and use getOrCreate(...) so exporters reuse the same HttpClient by default.

addresses the #789 issue 

### 2 Tests added 

`testHttpClientRegistryIsReused()` for the object equality check (re0use_ and concurrency test via `testHttpClientRegistryConcurrency()` In `OtlpHttpSpanExporterTest.kt`


RamachandraBhardwaj:
Hi @fractalwrench ! I have implemented this pattern where clients are reused based on the engine and timeout as a unique key. I've used concurrent map to prevent any race conditions too.

I noticed one part where `createDefaultHttpClient` was using a default timeout of 30_000 ms and the `EXPORT_REQUEST_TIMEOUT_MS` in `otlpHttpSpanExporter` is 10_000 ms.  I wanted to know if there is a reason for the difference or if it a known issue.

Please let me know if this addresses the issue properly or we require changes ! Thanks !

fractalwrench:
@RamachandraBhardwaj I don't think the discrepancy is intentional - 10s is the spec-default value so we should probably go with that.

Would you mind resolving the conflicts and then I can review the PR?

RamachandraBhardwaj:
@fractalwrench Thanks, I've set it 10_000 ms for the timeout and I've resolved the conflicts. Could you please take a look?

PS : I noticed 2 checks are failing because of a separate test failure in `ScopeImplTest.kt` in `implementation`. This is independent of my changes. Should I look into it as well?

fractalwrench:
The failure was existing on `main` - I've merged the latest in and that's resolved the build failure.

fractalwrench (approved):
LGTM - thanks!

PR #848, opened 2026-08-17
Read config from all sources for span limits

fractalwrench:
## Goal

Reads config from all sources (DSL, envars, yaml) for span limits. This is not hooked up to control the SDK yet, but now provides a good model of how config is intended to work. The envars/yaml are read and converted to `LogLimitsBehavior`, as is the DSL.

See #836 for additional details.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #849, opened 2026-08-17
Harden SpanModel and LogRecordModel

fractalwrench:
## Goal

Hardens the `SpanModel` and `LogRecordModel` classes by using `SdkErrorHandler` to guard potentially risky inputs.

## Testing

Added unit tests.


lemnik (approved):

bidetofevil (approved):
LGTM. Non-blocking general comment: I assume R8 will or even just the compiler will refactor out all those little string allocations so we pay no perf penalty when we use a no-op error handler?

PR #850, opened 2026-08-17
Implement attribute limit config reading

fractalwrench:
## Goal

Reads config from all sources (DSL, envars, yaml) for attribute limits. This is not hooked up to control the SDK yet, but now provides a good model of how config is intended to work. The envars/yaml are read and converted to `AttributeLimitsBehavior`, as is the DSL.

See https://github.com/open-telemetry/opentelemetry-kotlin/pull/836 for additional details.

## Testing
Added unit tests.


bidetofevil (approved):
LGTM

PR #881, opened 2026-08-18
Alter NoopSpanContext to use correct length IDs

### LINKED ISSUE #856: Alter NoopSpanContext to use correct length IDs
`NoopSpanContext` should use zeroed trace/span IDs of the correct length (32 vs 16 bytes).

Shaurya2k06:
## Goal

`NoopSpanContext` currently uses empty trace/span IDs (`""` / `ByteArray(0)`). The spec requires invalid IDs to be all-zero values of fixed length (16-byte trace ID, 8-byte span ID). This updates the noop context and id generator to match that, so consumers always see correctly sized IDs.

Fixes #856.

## Testing

Updated noop unit tests for hex and byte lengths. Verified with `:noop:jvmTest`, `:noop:jsTest`, and Detekt.


fractalwrench (approved):
LGTM - thanks!

PR #882, opened 2026-08-18
Replace getOrPut with alternative

### LINKED ISSUE #873: Replace getOrPut with alternative
`getOrPut` is non-atomic when the backing map is concurrent. Therefore we need to handle the race anywhere it happens within the codebase with something like the following instead, using `ApiProviderImpl` as an example:

```
fun getOrCreate(key: InstrumentationScopeInfo): T {
    map[key]?.let { return it }
    val candidate = supplier(key)
    return lock.write { map[key] ?: candidate.also { map[key] = it } }
}
```

Shaurya2k06:
## Goal

`getOrPut` is not atomic when the backing map is concurrent, so two callers can observe different cached Tracer/Meter/Logger (or context-key) instances. Replace those sites with a double-check and a write lock so the map winner is always returned, matching the pattern in #873 (`ApiProviderImpl` and the other production caches).

Fixes #873.

## Testing

Existing provider identity tests still assert the same instance is reused. Added a focused `ApiProviderImpl` unit test. Verified with `:implementation:jvmTest` and `:compat:jvmTest`.


fractalwrench:
Closing due to reasons explained on #873

PR #883, opened 2026-08-18
OTLP payloads carry one Resource + InstrumentationScope per span

### LINKED ISSUE #872: OTLP payloads carry one `Resource` + `InstrumentationScope` per span
OTLP payloads carry one `Resource` + `InstrumentationScope` per span. Grouping by resource/instrumentationScopeInfo before building the request would reduce the overall payload size. This affects both spans and logs.

Shaurya2k06:
## Goal

OTLP encode currently emits a Resource and InstrumentationScope envelope for every span/log. Group records by resource then instrumentation scope when building the protobuf request so a batch shares those envelopes and the payload is smaller. Applies to traces and logs.

Fixes #872.

## Testing

Added grouping cases for same resource/scope, same resource with different scopes, and different resources. Existing round-trip tests still pass. Verified with `:exporters-protobuf:jvmTest`.


exporters-protobuf/src/commonTest/kotlin/io/opentelemetry/kotlin/logging/export/ExportLogsServiceRequestTest.kt
```diff
@@ -142,4 +144,53 @@ class ExportLogsServiceRequestTest {
...
+        assertEquals(2, request.resource_logs[0].scope_logs.size)
+        assertEquals(1, request.resource_logs[0].scope_logs[0].log_records.size)
+        assertEquals(1, request.resource_logs[0].scope_logs[1].log_records.size)
+    }
+
+    @Test
+    fun testGroupsLogsWithDifferentResources() {
+        val request = listOf(
+            FakeLogRecordData(body = "a", resource = FakeResource()),
+            FakeLogRecordData(body = "b", resource = FakeResource()),
```
fractalwrench: Could these resources have different schemaUrls/attributes?

exporters-protobuf/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/export/ExportLogsServiceRequestCreator.kt
```diff
@@ -31,15 +31,16 @@ internal fun List<LogRecordData>.toExportLogsServiceRequest(): ExportLogsService
...
-    resource = resource.toProtobuf()
-)
-
-private fun LogRecordData.toScopeLogs(): ScopeLogs = ScopeLogs(
-    log_records = listOf(toProtobuf()),
-    scope = instrumentationScopeInfo.toProtobuf(),
-    schema_url = instrumentationScopeInfo.schemaUrl ?: ""
-)
+private fun List<LogRecordData>.toResourceLogs(): List<ResourceLogs> =
+    groupBy { it.resource }.map { (resource, logsForResource) ->
```
fractalwrench, line 35: I think grouping by resource like this requires `ResourceImpl` to implement equals/hashCode. Currently this will group by reference, but resource objects with the same attributes/schemaUrl might not match if they're two different objects.

fractalwrench (commented):
Thanks for opening this! I had a couple of comments but otherwise this looks like it's nearly ready to go.

Shaurya2k06:
I added equals/hashCode on ResourceImpl from attributes + schemaUrl so same-content resources group together. The different-resource tests now use distinct attributes/schemaUrls.

fractalwrench (approved):
LGTM - thanks!

PR #886, opened 2026-08-19
Read and combine config sources

fractalwrench:
## Goal

Reads config from the DSL, envars, and YAML in `OpenTelemetryConfigReader` and resolves them into `OpenTelemetryBehavior`, obeying precedence rules for DSL > YAML | envars.

This is not invoked from the implementation/compat modules but a subsequent PR will attempt to do this.

## Testing

Added unit tests.


bidetofevil (approved):
Nice. LGTM

PR #888, opened 2026-08-19
fix: accept only ASCII digits in hex validators

### LINKED ISSUE #868: Alter hexToByteArray() so it only accepts ASCII digits
`String.hexToByteArray()` uses Kotlin's Unicode-aware `Char.isDigit()` which accepts non-ASCII digits. We should assert with `c in '0'..'9'` instead to avoid potential crashes later on.

Shaurya2k06:
## Goal

Hex validators used `Char.isDigit()`, which accepts non-ASCII digits. That can make `isValidLowercaseHex()` succeed and later `toLong(16)` / `fromHex` throw. Check digits with `c in '0'..'9'` instead.

## Testing

Added HexTest cases for Arabic-Indic digits. Verified with `:implementation:jvmTest` (Hex*) and `:compat:jvmTest`.

Fixes #868


fractalwrench (approved):
LGTM - thanks!

PR #889, opened 2026-08-19
fix: do not inherit isRemote on locally created spans

### LINKED ISSUE #879: isRemote set erroneously for local spans with remote parent
isRemote is set erroneously for local spans with a remote parent in `TracerImpl`.

Shaurya2k06:
## Goal
Locally created spans were copying isRemote from a remote parent. Extracted parents stay remote; child SpanContexts are always local.

## Testing
Updated TracerInvalidIdTest. Ran implementation JVM tests for TracerInvalidIdTest and BuiltInSamplersTest.

Fixes #879


fractalwrench (approved):
LGTM - thanks!

PR #892, opened 2026-08-20
Make log snapshot internally consistent

### LINKED ISSUE #858: Return internally consistent LogRecordModel
`LogRecordModel` uses public accessors that acquire a lock each time. It should acquire the lock once and read backing files, in a similar way to how `Spanmodel.toSpanData()` works.

fractalwrench:
## Goal

Alters the log snapshotting so that it remains internally consistent similar to `SpanModel`. Closes #858


bidetofevil (approved):
LGTM

PR #893, opened 2026-08-20
Make MutableShutdownState and ScopeImpl use atomic operations

### LINKED ISSUE #874: Make MutableShutdownState and ScopeImpl use atomic operations
`AtomicLong` (or similar implementations) can be used to avoid check-then-set patterns in `MutableShutdownState` and `ScopeImpl`.

fractalwrench:
## Goal

Closes #874 by refactoring `MutableShutdownState` and `ScopeImpl` to use an `AtomicBoolean` operation. This prevents check-then-set races that the current implementation suffers from.


bidetofevil (approved):
LGTM. Too bad we have to have platform-specifics implementations.

platform-implementations/src/appleMain/kotlin/io/opentelemetry/kotlin/AtomicBoolean.kt
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.kotlin
+
+import platform.Foundation.NSLock
+
+public actual class AtomicBoolean actual constructor(initialValue: Boolean) {
```
breedx-splk, line 10: why can't we use the KMP `kotlin.concurrent.atomics.AtomicBoolean` ?

breedx-splk (approved):
Looks good to me, but not merging instantly because I had a question about the implementations of AtomicBoolean and I'm unsure why we can't use the KMP one. I suspect there's a good reason that I'm unaware of.

platform-implementations/src/appleMain/kotlin/io/opentelemetry/kotlin/AtomicBoolean.kt
fractalwrench: That's only available since Kotlin 2.1. We should move over to using it rather than hand-rolling when we drop support for Kotlin 2.0 https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.concurrent.atomics/-atomic-boolean/

PR #894, opened 2026-08-20
Allow shutting down SDK

### LINKED ISSUE #854: Alter createOpenTelemetry, createCompatOpenTelemetry, and NoopTelemetry return type
`createOpenTelemetry`, `createCompatOpenTelemetry`, and `NoopTelemetry` should have an `OpenTelemetrySdk` return type rather than `OpenTelemetry`. `OpenTelemetrySdk` should extend `TelemetryCloseable` so there's a simple way to shutdown the SDK and all its tracers/loggers.

This feels in the spirit of the spec given that the API module is unaffected, and it avoids an ugly cast that folks currently need to do.

fractalwrench:
## Goal

Closes #854 by allowing the SDK to shutdown, by making `OpenTelemetrySdk` implement `TelemetryCloseable`. This makes it much more practical for a library user to call `forceFlush` and `shutdown` in one place if they wish to stop the SDK.

I've avoided making this an explicit public API for now - folks still need to cast as this is technically an implementation detail according to the spec. We could revisit making this a public API in future.

## Testing

Added unit tests.

### DIFF
--- a/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/CompatOpenTelemetryImpl.kt
+++ b/compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/CompatOpenTelemetryImpl.kt
@@ -1,5 +1,10 @@
 package io.opentelemetry.kotlin
 
+import io.opentelemetry.kotlin.error.SdkErrorHandler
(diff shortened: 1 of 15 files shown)


bidetofevil (approved):
LGTM

PR #896, opened 2026-08-20
fix: derive SpanContextImpl.isValid from ID bytes

### LINKED ISSUE #878: Alter SpanContextImpl so that `isValid/isRemote` are derived from TraceFlags
`SpanContextImpl` currently sets `isValid/isRemote` separately from its `TraceFlags` property. This is confusing - for example, the invalid span context currently incorrectly states that it is sampled due to relying on the `TraceFlagsFactoryImpl` default.

Shaurya2k06:
## Goal
Derive SpanContextImpl.isValid from the trace/span ID bytes instead of a constructor flag. isRemote is unchanged.

## Testing
Added a SpanContextFactoryImplTest case for sampled flags with zero IDs. Ran implementation and compat JVM tests for SpanContext factories/tracer context.

Fixes #878


fractalwrench (approved):
LGTM - thanks!

PR #897, opened 2026-08-20
fix: do not throw on out-of-range OtelTraceState threshold

### LINKED ISSUE #876: Avoid require in OtelTraceState
`OtelTraceState.setThreshold` uses `require` which could throw `IllegalArgumentException` for an
invalid threshold.

Shaurya2k06:
## Goal
Stop OtelTraceState.setThreshold from throwing on an out-of-range threshold. Skip the write instead.

## Testing
Updated OtelTraceStateTest for negative and >14-hex-digit values. Ran :implementation:jvmTest for OtelTraceStateTest.

Fixes #876


fractalwrench (approved):
LGTM - thanks!

PR #898, opened 2026-08-20
Use I/O dispatcher for coroutines in processors/exporters

### LINKED ISSUE #863: Avoid holding lock on Dispatchers.Default in simple processors
`SimpleSpanProcessor` and `SimpleLogRecordProcessor` hold locks across I/O in `Dispatchers.Default`. We should avoid blocking this dispatcher and use alternatives instead.

fractalwrench:
## Goal

Closes #863 by using a suitable I/O dispatcher rather than `Dispatchers.default` (which could block the main thread) when exporting data.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #899, opened 2026-08-20
fix: deep-copy events, links, and attributes in toSpanData

### LINKED ISSUE #869: SpanModel collections have shallowed copy in toSpanData()
`SpanModel.toSpanData()` copies lists but not its elements for events/links/attributes, which means they can be mutated after a span is ended.

Shaurya2k06:
## Goal
Make SpanModel.toSpanData() deep-copy attributes/events/links so the snapshot cannot be mutated after a span ends.

## Testing
Added tests that mutate caller lists/byte arrays and assert the snapshot is unchanged. Ran :implementation:jvmTest for SpanData tests.

Fixes #869


fractalwrench (approved):
LGTM - thanks!

PR #900, opened 2026-08-20
fix: copy list and AnyValue attribute inputs on write

### LINKED ISSUE #862: AttributesModel should copy non-string lists
`setStringListAttribute` on `AttributesModel` copies the input so it cannot be mutated retroactively. We should do the same with `setBooleanListAttribute`, `setLongListAttribute`, `setDoubleListAttribute`, and `AnyValue`.

Shaurya2k06:
## Goal
Copy boolean/long/double lists and AnyValue payloads in AttributesModel so callers cannot mutate stored attributes after set.

## Testing
Added AttributesMutatorImplTest cases that mutate inputs after set. Ran :implementation:jvmTest for AttributesMutatorImplTest.

Fixes #862


fractalwrench (approved):
LGTM - thanks!

PR #903, opened 2026-08-21
fix: map StatusCode to protobuf with when, not ordinal

### LINKED ISSUE #859: Avoid mapping enum via ordinal
Avoid mapping an enum via ordinal in `SpanDataProtobufConversion.kt:34` - use `when` instead.

Shaurya2k06:
## Goal
Map span StatusCode to protobuf with when instead of enum ordinal so OTLP status does not depend on declaration order.

## Testing
Updated SpanDataProtobufConversionTest to assert proto StatusCode values and cover UNSET/OK/ERROR. Ran :exporters-protobuf:jvmTest for SpanDataProtobufConversionTest.

Fixes #859


fractalwrench (approved):
LGTM - thanks!

PR #910, opened 2026-08-21
fix: report SDK errors through Guard.kt

### LINKED ISSUE #866: Alter all invocations of onError to reportUserCodeError/Guard.kt
All invocations of `onError` should be altered to use functions within `Guard.kt` instead, as these have a try-catch guard for if the error handler throws.

Shaurya2k06:
## Goal
Report SDK errors through Guard.kt (reportError / reportUserCodeError / guard) so a throwing SdkErrorHandler cannot crash the host. ApiMisuse stays ApiMisuse.

## Testing
Added GuardTest coverage for reportError. Ran targeted sdk-common, implementation, and exporters-core jvmTest.

Fixes #866


fractalwrench (approved):
LGTM - thanks!

PR #920, opened 2026-08-21
fix: return snapshots from in-memory exporters

### LINKED ISSUE #864: InMemoryLogRecordExporter and InMemorySpanExporter defensive copy
InMemoryLogRecordExporter and InMemorySpanExporter should take a defensive copy when returning the collection, to avoid mutation of internal state.

henriquejsza:
## Goal

Return defensive snapshots from `exportedLogRecords` and `exportedSpans` instead of exposing their mutable backing lists. Add common regression tests proving that previously returned results do not change after later exports.

Closes #864

## Testing

- `./gradlew :exporters-in-memory:jvmTest :exporters-in-memory:jsNodeTest`
- `./gradlew :exporters-in-memory:allTests`
- `./gradlew :exporters-in-memory:apiCheck`
- `./gradlew detekt -x detektAndroidMain`, plus all non-Android `exporters-in-memory` Detekt source-set tasks

The full build and Android assembly were attempted, but the local SDK lacks accepted licenses for Android SDK 36. The same pre-compilation failure reproduces on an unmodified `upstream/main` worktree.


fractalwrench (approved):
LGTM - thanks!

PR #941, opened 2026-08-21
fix: enforce W3C baggage size limits on extract

### LINKED ISSUE #853: Enforce W3C Baggage limits on extract
`W3CBaggagePropagator` only enforces size limits when encoding, not on extract. We should alter `decode()` so that it applies the [same limits](https://www.w3.org/TR/baggage/#limits).

Shaurya2k06:
## Goal
Apply the same W3C baggage size limits on extract as on inject (4096 bytes per list-member, 8192 bytes total) so decode cannot accept unbounded headers.

## Testing
Added W3CBaggagePropagatorTest cases for oversized members and headers over 8192 bytes. Ran :implementation:jvmTest for W3CBaggagePropagatorTest.

Fixes #853


fractalwrench (commented):
I'm finding it a bit tricky to understand the `read()` function. I think it'd help to break it up into a couple of functions with expressive names, as that might make it clearer what each bit is doing.

fractalwrench (approved):
LGTM - thanks!

PR #943, opened 2026-08-21
fix: release span lock before processor.onEnd

### LINKED ISSUE #865: Reduce lock scope when ending span
`SpanModel.endInternal` holds a lock across `processor.onEnd()` - after the snapshot has been constructed, it's safe to invoke this without a lock.

Shaurya2k06:
## Goal
Release SpanModel’s write lock before SpanProcessor.onEnd so export is not serialized behind span mutation. Keep onEnding under the lock so processors can still mutate the span.

## Testing
Ran :implementation:jvmTest for SpanEndTest, SpanProcessorErrorHandlingTest, SpanModelTest, and the SpanProcess* onEnd/onEnding tests.

Fixes #865

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt
@@ -116,20 +116,21 @@ internal class SpanModel(
     @Suppress("NOTHING_TO_INLINE")
     private inline fun endInternal(timestamp: Long) {
         sdkErrorHandler.guard("Span.end failed") {
-            lock.write {
-                if (state == State.STARTED) {
-                    state = State.ENDING
-                    endTimestampImpl = timestamp
-                    sdkErrorHandler.guard {
-                        processor?.onEnding(ReadWriteSpanImpl(this))
-                    }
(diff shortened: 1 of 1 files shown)


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt
```diff
@@ -116,20 +116,21 @@ internal class SpanModel(
...
+            val snapshot = lock.write {
+                if (state != State.STARTED) return@write null
+                state = State.ENDING
+                endTimestampImpl = timestamp
+                sdkErrorHandler.guard {
+                    processor?.onEnding(ReadWriteSpanImpl(this))
+                }
+                state = State.ENDED
+                // Snapshot under the lock so processors retain plain data rather than this
+                // model. onEnd is invoked after release — the snapshot is immutable.
```
fractalwrench: Can you reinstate the original comment? Otherwise this looks good

Shaurya2k06: cool, works.

fractalwrench (approved):
LGTM - thanks!

PR #944, opened 2026-08-21
fix: ignore Unset when setting span status

### LINKED ISSUE #857: Setting an Unset status should be disallowed
The spec states that setting the Unset status should have no action: https://opentelemetry.io/docs/specs/otel/trace/api/#set-status

Shaurya2k06:
## Goal
Ignore Span.setStatus(Unset) so Unset cannot overwrite Error, matching the spec (Ok > Error > Unset).

## Testing
Updated SpanSimplePropertiesTest so Error then Unset stays Error. Ran :implementation:jvmTest for SpanSimplePropertiesTest.

Fixes #857

### DIFF
--- a/implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/tracing/SpanSimplePropertiesTest.kt
+++ b/implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/tracing/SpanSimplePropertiesTest.kt
@@ -113,6 +113,17 @@ internal class SpanSimplePropertiesTest {
     }
 
     @Test
+    fun testSpanStatusErrorCanBeOverriddenByOk() {
+        val span = tracer.startSpan("test")
+        span.setStatus(StatusData.Error("Whoops"))
+        span.setStatus(StatusData.Ok)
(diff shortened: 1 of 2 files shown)


implementation/src/commonTest/kotlin/io/opentelemetry/kotlin/tracing/SpanSimplePropertiesTest.kt
```diff
@@ -104,12 +104,12 @@ internal class SpanSimplePropertiesTest {
     }
 
     @Test
-    fun testSpanStatusErrorCanBeOverridden() {
+    fun testSpanStatusErrorCannotBeOverriddenByUnset() {
```
fractalwrench, line 107: I think an additional test case here that asserts `Error` can be overriden by `Ok` (but not vice-versa) would be helpful in catching regressions.

Shaurya2k06: done!

fractalwrench (approved):
LGTM - thanks!

Issue #784, opened 2026-08-11
TelemetryExporter does not close HttpClient


fractalwrench:
TelemetryExporter should close the HttpClient when shutdown is called.

Additionally, each `export` launches a coroutine that retains its batch for up to 8 attempts × 10-minute waits, causing memory to grow linearly if the collector isn't reachable. 

Also, Retry-After should be bounded to a sensible max upper limit.

breedx-splk:
> TelemetryExporter should close the HttpClient when shutdown is called.

I think that's only true when the exporters own the client. We have the public api overload [LogExportConfigDsl.otlpHttpLogRecordExporter](https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/export/OtlpLogRecordExporterApi.kt#L44)
which allows the user to pass in their own httpClient instance. In those cases, I don't think we should close their instance.

There can be cases where they want to shut down the exporter without shutting down their client.

fractalwrench:
I agree we shouldn't close a `HttpClient` if it's not our own.

The ideal behavior IMO is to share one `HttpClient` between exporters and close it down once no exporters are using it anymore (i.e. `shutdown()` has been invoked).

#928 is potentially relevant. To support all those configuration options we need a signature change, and I think it'd make sense to create a small DSL (which could more easily track whether a custom client was supplied)

PR #948, opened 2026-08-21
[FIX] : close http client on telemetry exporter shutdown + hard limits enforcement 

RamachandraBhardwaj:
## Goal

Add a separate suspend function inside of telemetry exporter to have `httpClient` closed . Added hard limits for `MAX_ATTEMPTS` and `MAX_RETRY_AFTER_MS` to prevent memory overload - when retrying. Addresses issue : #784 

## Testing

Test functions added to test `shutdownAction` in `TelemetryExporter.kt` and `OtlpHttpSpanExporterTest.kt`

### DIFF
--- a/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpClient.kt
+++ b/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpClient.kt
@@ -96,6 +96,10 @@ internal class OtlpClient(
         return header.toLongOrNull()?.takeIf { it >= 0 }?.let { it * 1000L }
     }
 
+    internal fun close() {
+        httpClient.close()
+    }
+
     private companion object {
         const val MAX_ERROR_BODY_BYTES: Long = 4 * 1024 * 1024
     }
--- a/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/TelemetryExporter.kt
+++ b/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/TelemetryExporter.kt
@@ -9,9 +9,12 @@ import kotlinx.coroutines.cancel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlin.coroutines.CoroutineContext
+import kotlin.math.min
 import kotlin.random.Random
 
 private const val SHUTDOWN_TIMEOUT_MS = 5000L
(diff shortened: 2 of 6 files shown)


RamachandraBhardwaj:
Hey @fractalwrench !

I just did a minor fix for otlpClient to have close method. Additionally for `MAX_ATTEMPTS` and `MAX_RETRY_AFTER_MS`, I'm using min of the constructor params and limits set by us. 
The assumption here is anyone using the client would give the maximum value as the names of the params suggest.

Please let me know if anything else could be added. Thanks !

exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpClient.kt
```diff
@@ -96,6 +96,10 @@ internal class OtlpClient(
         return header.toLongOrNull()?.takeIf { it >= 0 }?.let { it * 1000L }
     }
 
+    internal fun close() {
```
fractalwrench, line 99: This could probably use `AutoCloseable` instead?

exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/TelemetryExporter.kt
```diff
@@ -9,9 +9,12 @@ import kotlinx.coroutines.cancel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlin.coroutines.CoroutineContext
+import kotlin.math.min
 import kotlin.random.Random
 
 private const val SHUTDOWN_TIMEOUT_MS = 5000L
+private const val MAX_RETRY_AFTER_MS = 60_000L
```
fractalwrench, line 16: This seems a bit low - it wouldn't be desirable to bound retries to 1 minute as a server could plausibly be experiencing traffic and not want to retry for a longer period. A few hours feels like a reasonable upper bound as I'd assume we can mostly trust the server, and the max is just a backstop.

Similar comments for `MAX_ATTEMPTS`, perhaps we should increase that to something a bit higher?

breedx-splk (changes requested):
There are a couple of ways to create these exporters. In a couple cases, the httpClient instance has already been created by the user and is passed in directly. In those cases, I don't think that we would want to close it just because we're being closed. 

That seems like the responsibility of the thing that created the http client in the first place.

In the cases where the creation of the exporter makes its own instance, I think closing it is fine.

I put a comment about this in #784 so let's discuss there instead of in this PR?

PR #949, opened 2026-08-22
Remove unnecessary composite processor locks

### LINKED ISSUE #867: Remove unnecessary processor lock
`CompositeSpanProcessor` and `CompositeLogRecordProcessor` both use a lock, but this is unnecessary as `SpanProcessor` implementations are already required to be thread safe.

miachillgood:
## Goal

Fixes #867 by removing the read/write locks from the composite span and log record processors. Processor implementations are already required to be thread safe, so serializing every composite callback is unnecessary and prevents independent telemetry calls from running concurrently.

The processor iteration and error-handling behavior are unchanged.

## Testing

- `./gradlew :exporters-core:jvmTest --no-configuration-cache -Porg.gradle.java.installations.paths=<JDK_11_HOME>`
- `./gradlew :exporters-core:detektMetadataCommonMain --no-configuration-cache -Porg.gradle.java.installations.paths=<JDK_11_HOME>`


fractalwrench (approved):
LGTM - thanks!

Issue #808, opened 2026-08-12
Replace copies of collections in SpanModel and LogRecordModel with a read-only view


fractalwrench:
Replace copies of collections in `SpanModel` and `LogRecordModel` with a read-only view instead, as this reduces allocation churn.

PR #951, opened 2026-08-22
[FEAT] : Replace copies of collections in SpanModel and LogRecordModel with a read-only view  

RamachandraBhardwaj:
## Goal

Replace copies of collections in SpanModel and LogRecordModel with a read-only view - to prevent new copies. We added a separate package for reusable classes for views - `ReadOnlyListView` and `ReadOnlyMapView`. This addresses #808 

## Testing

added the tests for Read only views : `ReadOnlyMapViewTest.kt` and `ReadOnlyListViewTest.kt` to check functionality

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesModel.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/attributes/AttributesModel.kt
@@ -1,10 +1,14 @@
 package io.opentelemetry.kotlin.attributes
 
+import io.opentelemetry.kotlin.collections.ReadOnlyMapView
+
 internal class AttributesModel(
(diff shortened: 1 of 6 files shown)


RamachandraBhardwaj:
Hey @fractalwrench ! I've added a separate package for this so that helpers would be re-used across the repository. Please do let me know if this is a good pattern or something else can be changed. Thanks !

fractalwrench:
Thanks for attempting this @RamachandraBhardwaj. I've realized this approach (using read-only views) could cause a `ConcurrentModificationException` when altering a `SpanModel` in a `SpanProcessor`, so we likely _do_ need a copy. I'm therefore going to close this PR and the issue, but do feel free to pick up something else from the backlog. Sorry I didn't think of that before!

RamachandraBhardwaj:
Sure, no problem ! I'll look into other issues. Thank you !

Issue #952, opened 2026-08-22
Create a MetricStorage API


inetand:
## Goal
Define the internal MetricStorage API required by metric instruments to record measurements and by the SDK to read them back.

The API should provide the storage contracts needed for recording and collecting metric measurements without defining concrete instruments, aggregation implementations, or metric export.

## Scope

- Define readable and writable metric storage contracts.
- Associate storage with an instrument and its resolved stream descriptor.
- Support Long and Double measurements with attributes and context.
- Support recording timestamps and cumulative/delta collection semantics.
- Ensure recording is thread-safe.
- Provide an extensible storage boundary for future aggregation and exemplar support.
- Account for attribute cardinality and cardinality limits at the storage boundary where required by the SDK specification.

Concrete instruments, aggregations, readers, exporters, and provider wiring are out of scope.

## Specification
- [Metrics Data Model](https://opentelemetry.io/docs/specs/otel/metrics/data-model/)
- [Metrics API](https://opentelemetry.io/docs/specs/otel/metrics/api/)
- [Metrics SDK](https://opentelemetry.io/docs/specs/otel/metrics/sdk/)
- [Measurement Processing](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#measurement-processing)
- [Cardinality Limits](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#cardinality-limits)

colintheshots:
I can help here.

inetand:
@colintheshots 
Thank you for your support! I'll assign this to you here, then!

I think the right next steps would be:

1. Providing a no-op SynchronousMetricStorage and implementing both DeltaSynchronousMetricStorage and CumulativeSynchronousMetricStorage ([Aggregation temporality guidelines](https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/#synchronous-example))
2. Implementing something similar to [MetricStorageRegistry](https://github.com/open-telemetry/opentelemetry-java/blob/77b1ac4141b85759e5eee73bd06cc7091036b8af/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/MetricStorageRegistry.java) to register or reuse storage for resolved metric streams.

[...]

PR #953, opened 2026-08-22
Metrics API: Introduce data model types

inetand:
## GOAL

Introduce public interfaces representing the core OpenTelemetry Metrics Data Model to be used by MetricStorage.

Addresses this issue : [#952](https://github.com/open-telemetry/opentelemetry-kotlin/issues/952)

Added data types for Gauge, Sum, Histogram, and Exponential Histogram metric streams, including their corresponding point representations, aggregation temporality, timestamps, attributes, and histogram bucket data.

[The legacy Summary data model](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#summary-legacy) was intentionally excluded because it is not recommended for new OpenTelemetry applications.

MetricStorage and related API, including InstrumentDescriptors, Aggregation, AttributeProcessing, and View, will be introduced in the subsequent PRs.

## Testing

Compiled for all platforms. Detekt and apiCheck passed.

### DIFF
--- a/sdk-api/api/jvm/sdk-api.api
+++ b/sdk-api/api/jvm/sdk-api.api
@@ -271,6 +271,99 @@ public abstract interface class io/opentelemetry/kotlin/logging/model/ReadableLo
 	public abstract fun getTimestamp ()Ljava/lang/Long;
 }
 
+public final class io/opentelemetry/kotlin/metrics/data/AggregationTemporality : java/lang/Enum {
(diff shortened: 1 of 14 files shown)


fractalwrench (approved):
Thanks for opening this. I'm happy with this as a starting point for creating our implementation. I suspect that once a fuller implementation is in place we'll want to discuss and review the API in the SIG more fully, but IMO it'd be better to have something concrete there before doing that.

PR #955, opened 2026-08-23
Discard Span.setStatus calls with Unset status

### LINKED ISSUE #857: Setting an Unset status should be disallowed

sooraj-sky:
## Summary
- Per the [OTel trace API spec](https://opentelemetry.io/docs/specs/otel/trace/api/#set-status), an attempt to set span status to `Unset` SHOULD be ignored.
- `SpanModel.setStatus` previously allowed `Unset` to overwrite a previously-set `Error` status, which contradicts the spec's total order (`Ok > Error > Unset`).
- Added a guard so `Unset` is discarded, alongside the existing rule that `Ok` is final.
- Updated `SpanSimplePropertiesTest` to reflect the corrected behavior (renamed the test that previously asserted the wrong behavior, and added a coverage case for `Error` being overridden by `Ok`).

Fixes #857

## Test plan
- [x] `./gradlew :implementation:jvmTest` — all tests pass, including new/updated status tests
- [x] `./gradlew detekt` — no lint violations

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt
@@ -94,7 +94,8 @@ internal class SpanModel(
 
     override fun setStatus(status: StatusData) {
         mutate("Span.setStatus failed") {
-            if (statusImpl !is StatusData.Ok) {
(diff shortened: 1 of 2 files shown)


fractalwrench:
Thanks for attempting this @sooraj-sky but #944 got there first. You'd be very welcome to pick up another issue in the repo.

PR #960, opened 2026-08-24
fix: validate instrument name syntax (#931)

### LINKED ISSUE #931: Validate instrument name
## Instrument name syntax validation

Ensure that instrument names match the rules specified here: https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-name-syntax

Shaurya2k06:
## Goal
Validate instrument names against the spec ABNF (ASCII, start with a letter, max 255, `_` `.` `-` `/`). Invalid names report ApiMisuse and return a noop instrument so exporters never see illegal names and the host is not destabilized.

## Testing
Added InstrumentNameTest for valid/invalid syntax and length. Added an integration test on UpDownCounterImplTest verifying MeterImpl returns a disabled noop and reports ApiMisuse. Ran :implementation:jvmTest, detekt, and apiCheck.

Fixes #931


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/InstrumentName.kt
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Checks [name] against the
+ * [instrument name syntax](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-name-syntax):
+ * a non-empty ASCII string starting with a letter, followed by up to
+ * [MAX_INSTRUMENT_NAME_CHARS] - 1 alphanumeric characters, `_`, `.`, `-`, or `/`.
+ *
+ * Invalid names are reported as [SdkError.ApiMisuse] so callers can fall back to a noop
+ * instrument rather than destabilizing the host or risking a name collision from rewriting.
+ */
+internal fun SdkErrorHandler.isValidInstrumentName(name: String): Boolean {
```
fractalwrench: I'd prefer this not to be an extension function on `SdkErrorHandler` as it makes the symbol available to every class in the module. It'd be preferable to have it on `MeterImpl` and take `SdkErrorHandler` as a param instead.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/InstrumentName.kt
```diff
@@ -0,0 +1,39 @@
...
+    if (INSTRUMENT_NAME.matches(name)) {
+        return true
+    }
+    reportError(
+        SdkError.ApiMisuse(
+            api = "Instrument.name",
+            message = "Instrument name \"$name\" is invalid; returning a noop instrument. " +
```
fractalwrench: nit: don't include `$name` as this explodes the error message cardinality

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/MeterImpl.kt
```diff
@@ -1,15 +1,24 @@
...
     override fun createDoubleUpDownCounter(
         name: String,
         unit: String?,
         description: String?,
-    ): DoubleUpDownCounter = DoubleUpDownCounterImpl(name, unit, description)
+    ): DoubleUpDownCounter {
+        if (!sdkErrorHandler.isValidInstrumentName(name)) {
+            return NoopOpenTelemetry.meterProvider.getMeter("")
+                .createDoubleUpDownCounter(name, unit, description)
+        }
```
fractalwrench: IMO it'd be nice to have syntax along these lines so that it's easy to get the same logic for other instruments when they're added, without having to repeat the checks:


```kotlin
obtainMeter(name).createDoubleUpDownCounter(name, unit, description)

private fun obtainMeter(name: String): Meter {
    return if (!sdkErrorHandler.isValidInstrumentName(name)) {
        NoopOpenTelemetry.meterProvider.getMeter("")
    } else {
        this
    }
}
```

fractalwrench (commented):
Left a few comments inline but this looks like it's nearly there.

Shaurya2k06:
Went with a slightly different shape, obtainNoopMeterIfInvalid(name) returns the noop meter when invalid, null when valid, since returning this from obtainMeter and calling .createDoubleUpDownCounter on it recurses back into the same override forever. This keeps the "check once, reuse across instrument types" idea without the infinite loop. 

Let me know if you have another plan of action in mind.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/InstrumentName.kt
Shaurya2k06: addressed.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/InstrumentName.kt
Shaurya2k06: fixed.

fractalwrench (approved):
LGTM - thanks!

Issue #933, opened 2026-08-21
Instrument description validation


fractalwrench:
The Instrument description parameter should follow the validation rules described here: https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-description

Shaurya2k06:
Hey @fractalwrench , this one seems the easiest out of the issues you opened a couple of days ago. Could it be assigned to me?

PR #961, opened 2026-08-24
fix: sanitize instrument units to spec (#932)

### LINKED ISSUE #932: Instrument unit validation
Ensure the instrument unit parameter matches the rules specified here: https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-unit

Shaurya2k06:
## Goal
Sanitize instrument units to the spec: optional case-sensitive ASCII, at most 63 characters. Invalid units report ApiMisuse and are dropped (null) so the host is not destabilized and exporters do not see illegal units. The instrument itself is still created.

## Testing
Added InstrumentUnitTest for null, empty, 63, 64, and non-ASCII. Added an integration test on UpDownCounterImplTest verifying MeterImpl drops the unit and reports ApiMisuse. Ran :implementation:jvmTest, detekt, and apiCheck.

Fixes #932


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/InstrumentUnit.kt
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Sanitizes [unit] to satisfy the
+ * [instrument unit](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-unit)
+ * requirements: a case-sensitive ASCII string of at most [MAX_INSTRUMENT_UNIT_CHARS] characters.
+ *
+ * Invalid units are reported as [SdkError.ApiMisuse] and dropped to `null` rather than
+ * destabilizing the host or being rewritten to a misleading value. `null` stays `null`.
+ */
+internal fun SdkErrorHandler.sanitizeInstrumentUnit(unit: String?): String? {
```
fractalwrench: I'd prefer this to be a private function on `MeterImpl` that passes in `SdkErrorHandler`. That avoids namespace pollution for other classes in this module that don't need to know these implementation details.

Shaurya2k06: done, same way as #933

fractalwrench:
@Shaurya2k06 can you rebase this please? Thanks

fractalwrench (approved):
LGTM - thanks!

PR #971, opened 2026-08-24
Alter default propagator to W3cBaggage and TraceContext

fractalwrench:
## Goal

Alters the SDK's default propagator to W3cBaggage and TraceContext, as written in the spec. Previously the default was for no propagators to be configured which resulted in the SDK dropping trace context/baggage when the SDK was enabled.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. Just to confirm: the new behaviour is that the default SDK propagators are provided, which should just be the no-op ones (whether they "op" by default is still being discussed)?

Issue #954, opened 2026-08-22
Create an Aggregation API


inetand:
## Goal
Define internal Aggregation and View APIs for configuring metric processing and resolving metric streams.

## Scope
- Define aggregation types and their configuration.
- Define instrument selection by name, kind, unit, and instrumentation scope.
- Configure stream name, description, attributes, aggregation, and cardinality limit.
- Resolve matching Views independently into stream configurations consumed by MetricStorage.
- Preserve defaults from the instrument and MetricReader when View values are omitted.

Concrete exporters implementation is out of scope.

## Specification
- [Aggregation](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#aggregation)
- [View](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view)
- [Stream Configuration](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#stream-configuration)
- [Measurement Processing](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#measurement-processing)

PR #976, opened 2026-08-24
Introduce internal Metrics API foundation

inetand:
## GOAL

Created the internal Metrics SDK foundations for configuring and storing metric streams. Addresses https://github.com/open-telemetry/opentelemetry-kotlin/issues/952 and https://github.com/open-telemetry/opentelemetry-kotlin/issues/954

Added Aggregation and View contracts, instrument and metric descriptors, attribute processing, aggregation resolution, and readable/writable MetricStorage APIs.

Concrete aggregation and instrument implementations will be added in subsequent PRs.

## Testing

Compiled for all platforms. Detekt, apiCheck, and Android test compilation passed.

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/MetricDescriptor.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/MetricDescriptor.kt
@@ -1,0 +1,63 @@
+package io.opentelemetry.kotlin.metrics
+
+import io.opentelemetry.kotlin.metrics.instrument.InstrumentDescriptor
+import io.opentelemetry.kotlin.metrics.view.AttributesProcessor
+import io.opentelemetry.kotlin.metrics.view.View
+
+/**
+ * SDK configuration for a metric stream after applying an instrument-matching View.
+ *
+ * One instrument may produce multiple independently configured streams when multiple Views match.
+ *
+ * https://opentelemetry.io/docs/specs/otel/metrics/sdk/#measurement-processing
+ */
+internal class MetricDescriptor private constructor(
+    val name: String,
+    val description: String?,
+    val sourceInstrument: InstrumentDescriptor,
+    val view: View,
+    val attributesProcessor: AttributesProcessor,
+    val cardinalityLimit: Int,
+) {
+    val unit: String? = sourceInstrument.unit
+
+    override fun equals(other: Any?): Boolean =
+        this === other ||
+            other is MetricDescriptor &&
+            name.equals(other.name, ignoreCase = true) &&
+            description == other.description &&
+            sourceInstrument == other.sourceInstrument &&
+            view.aggregation == other.view.aggregation &&
+            attributesProcessor == other.attributesProcessor &&
+            cardinalityLimit == other.cardinalityLimit
+
+    override fun hashCode(): Int {
+        var result = name.lowercase().hashCode()
+        result = 31 * result + description.hashCode()
+        result = 31 * result + sourceInstrument.hashCode()
+        result = 31 * result + view.aggregation.hashCode()
+        result = 31 * result + attributesProcessor.hashCode()
+        result = 31 * result + cardinalityLimit
+        return result
+    }
+
+    override fun toString(): String =
+        "MetricDescriptor(name=$name, description=$description, sourceInstrument=$sourceInstrument, " +
+            "view=$view, cardinalityLimit=$cardinalityLimit)"
+
+    companion object {
+        fun create(
+            sourceInstrument: InstrumentDescriptor,
+            view: View,
+            attributesProcessor: AttributesProcessor = view.attributesProcessor,
+            cardinalityLimit: Int = view.cardinalityLimit,
+        ): MetricDescriptor = MetricDescriptor(
+            name = view.name ?: sourceInstrument.name,
+            description = view.description ?: sourceInstrument.description,
+            sourceInstrument = sourceInstrument,
+            view = view,
+            attributesProcessor = attributesProcessor,
+            cardinalityLimit = cardinalityLimit,
+        )
+    }
+}
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/MetricDescriptor.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/MetricDescriptor.kt
@@ -5,9 +5,15 @@ import io.opentelemetry.kotlin.metrics.view.AttributesProcessor
 import io.opentelemetry.kotlin.metrics.view.View
 
 /**
- * SDK configuration for a metric stream after applying an instrument-matching View.
+ * Resolved SDK configuration for a metric stream after applying an instrument-matching View.
  *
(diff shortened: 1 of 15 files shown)


inetand:
#### Intended usage of SdkInstrument

Concrete instruments implement both their public API and SdkInstrument. The latter supplies descriptor-backed instrument metadata shared by SDK processing like this:

```kotlin
internal class DoubleHistogramImpl(
...
) : DoubleHistogram, SdkInstrument 
```

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/instrument/InstrumentDescriptor.kt
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Identifying parameters captured when a Meter creates an instrument.
+ *
+ * Instrument names are compared case-insensitively. Kind, unit, description, and numeric value
+ * type are identifying; [advice] is not, so the first advice registered for otherwise identical
+ * instruments is retained by the Meter.
+ *
+ * https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument
+ */
+internal class InstrumentDescriptor(
```
fractalwrench, line 12: Could this be a data class instead to avoid the manual equals/hashcode implementations?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/MetricDescriptor.kt
```diff
@@ -0,0 +1,63 @@
...
+import io.opentelemetry.kotlin.metrics.view.View
+
+/**
+ * SDK configuration for a metric stream after applying an instrument-matching View.
+ *
+ * One instrument may produce multiple independently configured streams when multiple Views match.
+ *
+ * https://opentelemetry.io/docs/specs/otel/metrics/sdk/#measurement-processing
+ */
+internal class MetricDescriptor private constructor(
```
fractalwrench, line 14: Could this be a data class?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/view/View.kt
```diff
@@ -0,0 +1,26 @@
...
+internal data class View(
+    val selector: InstrumentSelector,
+    val name: String? = null,
+    val description: String? = null,
+    val aggregation: Aggregation? = null,
+    val attributesProcessor: AttributesProcessor = AttributesProcessor.Noop,
+    val cardinalityLimit: Int = MetricStorage.DEFAULT_MAX_CARDINALITY,
+) {
+    init {
+        require(cardinalityLimit > 0) { "cardinalityLimit must be positive" }
```
fractalwrench, line 24: As a general rule the preference should be to degrade on invalid input rather than throwing exceptions, given that this SDK could be included in other apps and an exception could crash it, thus losing telemetry. I'd suggest just defaulting to `MetricStorage.DEFAULT_MAX_CARDINALITY` if the supplied limit is invalid?

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/MetricStorage.kt
```diff
@@ -0,0 +1,35 @@
...
+     * Delta storage advances its interval after collection; cumulative storage retains its start
+     * timestamp and previously observed synchronous series.
+     */
+    fun collect(
+        resource: Resource,
+        instrumentationScopeInfo: InstrumentationScopeInfo,
+        timestampEpochNanos: Long,
+    ): MetricData
+
+    fun setEnabled(enabled: Boolean)
```
fractalwrench, line 30: small nit: consider `var enabled: Boolean`

fractalwrench (commented):
Thanks for tackling this. I left a few comments inline but it feels like this is nearly there.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/MetricDescriptor.kt
inetand: Yeah, since it’s an internal class, I think it’s safe to make it a data class, but we’d still need custom equals/hashCode implementations due to the case-insensitive name matching and the fact that equality is based only on the resolved aggregation from View (not its selector or unresolved fields).

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/instrument/InstrumentDescriptor.kt
inetand: Yeah, that's pretty unclear due to vague documentation: 

> Kind, unit, description, and numeric value type are identifying; [advice] is not, so the first advice registered for otherwise identical instruments is retained by the Meter.

What I really meant here is that advice isn't used to identify `InstrumentDescriptors`.

That said, it can be a data class, but since we need to exclude advice from equals/hashCode and implement case-insensitive name matching, we’d still need to implement those manually.

I can make the class itself a data class and rephrase the docs

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/MetricDescriptor.kt
fractalwrench: Happy to leave this as-is, given that I missed the case-insensitive matching.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/instrument/InstrumentDescriptor.kt
fractalwrench: Happy to avoid making this a data class and rephrase the docs/leave a comment stating why the equals implementation is different from what a generated version would provide

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/metrics/view/View.kt
inetand: @fractalwrench That’s a good point. I agree that the SDK should be more resilient when invalid data is passed, rather than potentially causing the client app to crash.

So, I’ve tried a few approaches, including constructor-level fallback and a `fun View(...): View` factory method, but this is a bit tricky with a data class. The property needs to be a val, and making the primary constructor private would also affect the generated copy() method.

Given those constraints, I’m leaning toward documenting the valid range and validating the value on the caller side (inside SDK). Does this approach make sense?

fractalwrench: @inetand I assume that means the SDK entrypoint would validate the range is valid, and no-op/warn in a way that avoids constructing this class? That would sound reasonable to me

inetand: Yes, exactly. I meant validating on the entrypoint side, and warn + falling back to the default value when creating the View, as you suggested above

fractalwrench (approved):
LGTM - thanks!

Issue #924, opened 2026-08-21
Configuration: ConsoleExporter


fractalwrench:
## ConsoleExporter

ConsoleExporter: writes telemetry to standard output for debugging.
Contains the following fields: none — selecting the exporter is the whole configuration

Declarative config path: `.{tracer_provider,logger_provider}.processors[].*.exporter.console`
Schema type: `ConsoleExporter`
Env vars: `OTEL_TRACES_EXPORTER=console`, `OTEL_LOGS_EXPORTER=console`

Spec links:

- Stdout span exporter: https://opentelemetry.io/docs/specs/otel/trace/sdk_exporters/stdout/
- Stdout log exporter: https://opentelemetry.io/docs/specs/otel/logs/sdk_exporters/stdout/
- Exporter selection: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#exporter-selection
- YAML schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/common.yaml

Implementation checklist:

- [ ] Behavior implemented in `behavior` module?
- [ ] Readable from `config-dsl` module?
- [ ] Readable from `config-envar` module?
- [ ] Readable from `config-yaml` module?
- [ ] Compat mode SDK obeys config
- [ ] Implementation mode SDK obeys config

RabahSend:
Hi, I'd like to work on this issue. Is it available to take?

fractalwrench:
@RabahSend you're welcome to attempt this. I'd suggest splitting the work up into multiple PRs, and you should also be aware that config isn't read properly until #886 lands. I also don't have a pre-existing example of how the SDK should obey config, but hopefully that should land sometime next week.

PR #978, opened 2026-08-24
Add console exporter behavior (#924)

RabahSend:
## Goal

Puts console exporter selection on `LogRecordProcessorBehavior` and `SpanProcessorBehavior` (not on the providers). Rebased onto #996 so providers hold `processor`; `console` is a field on that processor.

`ConsoleExporterBehavior` is still empty: presence means console was selected. `null` on the processor means it was not. Merge uses `mergeNode`.

This PR is still behavior only. It does not read console from DSL, env, or YAML, and it does not make the SDK create stdout exporters.

## Testing

Added unit tests to `TracerProviderBehaviorTest` and `LoggerProviderBehaviorTest` covering:
- the default unset state;
- merging when only one layer sets console;
- merging when neither layer sets it.

Validated with:

`./gradlew :behavior:jvmTest`


fractalwrench (commented):
It wasn't very clear from what was written on the issue, but semantically I think it would make sense as an ultimate goal for the `LoggerProviderBehavior/TracerProviderBehavior` classes to hold a `LogRecordProcessor/SpanProcessor` property, that is configured either via the DSL, yaml, or envar. I think there are a few extra steps to get to that point so am not too sure how to proceed this PR. Perhaps removing the `console` property from `LoggerProviderBehavior` for now would be a way forward?

RabahSend:
Thanks, that matches the YAML tree (provider -> processor -> exporter -> console). I put console on the provider to keep this PR small.

I’d rather drop console from both TracerProviderBehavior and LoggerProviderBehavior in this PR, and keep only ConsoleExporterBehavior, until processor behavior exists. Does that work, or would you prefer to close this and wait?

fractalwrench:
@RabahSend my gut feel is that it'd make sense to use a top-down approach - i.e. move specifying a `LogRecordProcessor` onto `LoggerProviderBehavior`, and the same for spans. The high-level ordering I'd go with for this feature would roughly follow these lines:

1. `Move `LogRecordProcessor` onto `LoggerProviderBehavior` (from existing DSL)
2. Do the same for spans
3. Setup YAML parsing of processors/exporters, with ConsoleExporter as the MVP that proves the approach (i.e. rebase this PR)
4. Same as (3), but for envar config (if this is even written in the spec)
5. Support other processor/exporter options until the spec is fully met

There's a bit of uncertainty/wiggle room as I may have missed bits from the spec, but that's my current thinking

RabahSend:
Thanks, that ordering makes sense.

I’ll follow the top-down approach: first put LogRecordProcessor on LoggerProviderBehavior from the existing DSL, then the same for spans. I’ll rebase this ConsoleExporter PR once that processor/exporter shape exists, with console as the MVP exporter.

I’ll start the logger-processor work on a separate branch from main and leave this PR as-is until then.

fractalwrench:
@RabahSend can you rebase now that #996 has landed?

RabahSend:
@fractalwrench Rebased on the latest changes, and all checks are passing now. Thanks!

behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/LogRecordProcessorBehavior.kt
```diff
@@ -8,11 +8,16 @@ import io.opentelemetry.kotlin.ExperimentalApi
...
  */
 @ExperimentalApi
-class LogRecordProcessorBehavior : Behavior<LogRecordProcessorBehavior> {
+data class LogRecordProcessorBehavior(
 
-    override fun mergeWith(higher: LogRecordProcessorBehavior): LogRecordProcessorBehavior = higher
+    /**
+     * Console log exporter. Selecting it is the whole configuration.
+     */
+    val console: ConsoleExporterBehavior? = null,
```
fractalwrench, line 16: I'm finding it hard to visualise how this class would be used by the implementation/compat modules, and whether exposing this type to the `implementation/compat` modules would be the best approach or whether there's a more general approach. Would you mind adding a comment reply that sketches out how you'd anticipate this API being used? Thanks

behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/ConsoleExporterBehavior.kt
```diff
@@ -0,0 +1,19 @@
...
+import io.opentelemetry.kotlin.ExperimentalApi
+
+/**
+ * Selecting the console exporter. This type has no fields: choosing it is the whole configuration.
+ *
+ * https://opentelemetry.io/docs/specs/otel/trace/sdk_exporters/stdout/
+ * https://opentelemetry.io/docs/specs/otel/logs/sdk_exporters/stdout/
+ */
+@ExperimentalApi
+class ConsoleExporterBehavior : Behavior<ConsoleExporterBehavior> {
```
CLOVIS-AI: If it is stateless, why not use an `object`? `data object` gives `hashCode` and `equals`.

behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/LogRecordProcessorBehavior.kt
RabahSend: Yeah, this PR still only stores the intent. `processor.console != null` means "console was selected". Same as #996, nothing in `implementation`/`compat` reads this yet.

The way I'd use it later: after config is merged into one `OpenTelemetryBehavior`, those modules just look at that flag and build a real processor:

```kotlin
if (behavior.loggerProvider?.processor?.console != null) {
    simpleLogRecordProcessor(stdoutLogRecordExporter())
}
```

Same idea for traces with `simpleSpanProcessor(stdoutSpanExporter())`. Compat would do the Java equivalent. The providers already know how to attach a live processor, we just aren't creating it from behavior yet.

I'm not sure `ConsoleExporterBehavior` is what we want to leak into those modules long term. The YAML tree is `processors[].simple.exporter.console`, and this flattens that to `processor.console`. Fine as an MVP for console, but we might want a more general exporter node when OTLP shows up. Happy to nest it like the schema now if you prefer that.

fractalwrench (approved):
LGTM - thanks! We'll see how this looks from the `Behavior` interface side once a few more parts are in, and then can adjust as needed...

Issue #851, opened 2026-08-18
Decide on default behavior for Propagator API


fractalwrench:
## Current state

Currently `:api` exposes the `TextMapPropagator`, `TextMapGetter`, and `TextMapSetter` interfaces, along with a `OpenTelemetry.propagator`. Concrete propagators such as `W3CBaggagePropagator` are internal and can only be obtained via the `sdk-api` during SDK initialization.

If the app never configures with `propagator {}` during init, a no-op propagator is used. This means that a fully initialized SDK would drop trace context on every remote call.

Instrumentation should only depend on the `api` module, and therefore it can access the configured propagator as a read-only property. By default, this would be a no-op instance, and it's not possible for instrumentation to construct or set propagators.

## What the spec requires

I picked the requirements from the [spec](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#global-propagators) that seem relevant for our discussion:

1. The OpenTelemetry API MUST provide a way to obtain a propagator for each supported Propagator type.
2. The OpenTelemetry API MUST use no-op propagators unless explicitly configured otherwise.
3. Instrumentation libraries SHOULD call propagators to extract and inject the context on all remote calls. 
4. Propagators, depending on the language, MAY be set up using various dependency injection techniques or available as global accessors.
5. Implementations MUST offer a facility to group multiple `Propagator`s from different cross-cutting concerns in order to leverage them as a single entity. (achieved in sdk-api, but not api)
6. Propagators must be distributed in [core packages](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#propagators-distribution) (seems to contradict requirement 1)

(2) is currently fulfilled by opentelemetry-kotlin. All other requirements in that list are not fully met.

## What use-case makes the spec require this?

My understanding is that allowing instrumentation to construct a propagator helps keep the trace context alive. For example, if a server uses a [W3C TraceContext header](https://www.w3.org/TR/trace-context/) for a request but the client does not, it will not be possible to attach the two traces together and view a distributed trace.

An additional use-case would be baggage propagation. For example, if opentelemetry-kotlin is running on a server environment and receives a call with baggage and then makes a call to another server, then a no-op propagator would drop the baggage.

## Client/mobile concerns

My main concern is that if a developer switched something off, it should stay switched off. Particularly in a mobile environment, disabling an SDK brings the expectation that it's completely disabled, and it would be negatively received if propagation contributed to crashes/performance degradation with no way to switch the behavior off.

[...]

PR #979, opened 2026-08-25
[POC] Make Propagators available when SDK is not initialized

fractalwrench:
## Goal

POC that demonstrates how the project could structure the codebase to fulfil the requirement of #851 that propagators are accessible independent of the SDK. This is a proof-of-concept that is intended to show the general approach, and isn't a final implementation. There are a few key decisions made:

1. A new `propagators` module is created and contains functions to create working implementations of propagators (`Propagators.create().w3cBaggagePropagator()). The module depends on `api` but is separate from it
2. The context/context storage implementation is similarly extracted to a separate module (these could be combined into the same module if desired)
3. The `implementation` and `compat` layers both use the KMP propagators. Functionally, that means anybody using the compatibility mode will now use the KMP propagators rather than opentelemetry-java's implementation, which feels a reasonable-ish tradeoff to meet the requirement
4. I've added an optional flag to disable the use of propagators in this way. I think this would have utility on mobile given folks often want an SDK to be disabled completely

I will attempt to break this up into smaller PRs that are easier to review (assuming folks are happy with the general direction).


bidetofevil:
Overall, I still don't feel right about the default behaviour of the no-op implementation to propagate. I understand that on some platforms, existing norms, compatibility, and unique use cases have forced the hand and made this a thing, but conceptually, a no-op that actually ops doesn't seem like a reasonable requirement to enforce at the API level as default behavior.

This is especially sensitive for an SDK whose primary use case right now is for mobile apps, where release that contain this code live in the wild forever and changes cannot be rolled back. The trust in the SDK to be no-op is of paramount importance when you have no recourse post release, so I would very much like to NOT have propagation on by default.

I can understand having this as an optional, opt-in configuration, but I would hate to put this in just to satisfy a requirement on paper that arguably does more harm than good for current -- one that isn't expected to boot, unless you have an intimate known of the OTel Propagators API.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/init/OpenTelemetryConfigImpl.kt
```diff
@@ -68,6 +71,12 @@ internal class OpenTelemetryConfigImpl(
         propagatorCfg.action()
     }
 
+    override var instrumentationPropagation: Boolean = true
```
bidetofevil, line 74: I think we should set this to `false` by default (and consider doing the same on the compat impl too). More reasoning in the other comments.

propagators/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/Propagators.kt
```diff
@@ -0,0 +1,84 @@
...
+    public fun setEnabled(enabled: Boolean) {
+        this.enabled = enabled
+    }
+
+    /**
+     * Returns the propagator implementations without the [setEnabled] gate applied, so that a
+     * propagator the application configured explicitly always propagates.
+     */
+    @InternalPropagatorApi
+    public fun unrestricted(): PropagatorApi = PropagatorApiImpl
```
bidetofevil, line 45: Do we actually need this? `GatedPropagatorApi` should be able to handle this if `enabled` is set correctly?

propagators/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+```kotlin
+val propagator = Propagators.create().w3cBaggage()
+propagator.inject(context, headers, setter)
+```
+
+No SDK instance or object factory is needed: a propagator operates on whichever object
+implementations belong to the `Context` it is given, so one instance is safe to share across
+backends. `Context` and `Baggage` keep working when no SDK is installed (see `context-impl`), which
+means instrumentation can keep a trace alive across a remote call even if the application never
```
bidetofevil, line 17: This being the use case is why I don't think we need this behaviour as a default-on for this SDK, which defaults to mobile expectations - and I don't expect that it's common for mobile apps to be a hop in the middle of the distributed trace.

sdk-api/src/commonMain/kotlin/io/opentelemetry/kotlin/init/OpenTelemetryConfigDsl.kt
```diff
@@ -51,6 +51,21 @@ public interface OpenTelemetryConfigDsl : ResourceConfigDsl, ConfigFileDsl {
...
+     *
+     * Set this to `false` if the application must be able to guarantee that no context leaves the
+     * process — for example where baggage could carry data the user has not consented to sharing.
+     * It has no effect on the propagator configured via [propagator], which always propagates.
+     *
+     * This switch is process-wide rather than scoped to this instance, because the propagators it
+     * governs are obtained without an [io.opentelemetry.kotlin.OpenTelemetry] instance. The last
+     * SDK to set it wins.
+     */
+    public var instrumentationPropagation: Boolean
```
bidetofevil, line 67: Consider rename this to something that makes it clear that a no-op SDK implementation will actually still always propagate (e.g. `alwaysPropagate`)

bidetofevil (commented):
No issues with the code moving around to enable propagation even for a np-op SDK impl, but I have strong reservations about enabling that behaviour by default. See previous comment.

PR #980, opened 2026-08-25
fix: make DefaultImplicitContextStorage.current volatile

### LINKED ISSUE #962: Default implicit-context storage has an unsynchronised mutable field
`DefaultImplicitContextStorage` holds the current context in a plain `var`. This should use `@Volatile` instead.

Shaurya2k06:
## Goal
Make GLOBAL implicit-context storage visible across threads by marking
DefaultImplicitContextStorage.current @ Volatile. A plain var can leave
another thread seeing a stale context.

## Testing
Existing DefaultImplicitContextStorageImplTest covers set/get. Ran
implementation jvmTest for context storage/scope plus detekt.

Fixes #962


fractalwrench (approved):
LGTM - thanks!

PR #981, opened 2026-08-25
fix: construct SpanContext from bytes without a hex round trip

### LINKED ISSUE #968: ByteArray variant of SpanContextFactoryImpl create() has a hex string round trip
The `ByteArray` variant of `SpanContextFactoryImpl` `create()` has a hex string round trip:

```kotlin
override fun create(traceIdBytes: ByteArray, spanIdBytes: ByteArray, ...) = create(
    traceIdBytes.toHexString(), spanIdBytes.toHexString(), ...
)
```

The `String` overload then validates and calls `hexToByteArray()` to get back to bytes. We should alter the implementation so that providing a ByteArray results in no extra allocations, and providing a hex string calls the ByteArray overload.

Shaurya2k06:
## Goal
Stop SpanContextFactoryImpl's ByteArray create() from round-tripping through
hex. Bytes construct SpanContextImpl directly; hex strings convert with
hexToByteArray and delegate. Invalid IDs still become the generator's
all-zero IDs.

## Testing
Existing string create tests still cover valid/invalid/uppercase hex.
Added ByteArray cases for valid, all-zero, and wrong-length IDs. Ran
implementation jvmTest for SpanContextFactoryImpl / SpanContextImpl /
TracerInvalidIdTest plus detekt.

Fixes #968

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/SpanContextFactoryImpl.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/factory/SpanContextFactoryImpl.kt
@@ -27,24 +27,13 @@ internal class SpanContextFactoryImpl(
         traceFlags: TraceFlags,
         traceState: TraceState,
         isRemote: Boolean,
-    ): SpanContext {
-        val isValidTraceId = isValidTraceId(traceId)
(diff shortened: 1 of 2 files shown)


fractalwrench (approved):
LGTM - thanks!

PR #982, opened 2026-08-25
fix: match span event bridge IDs by bytes, not hex

### LINKED ISSUE #969: `SpanEventBridgeLogRecordProcessor` compares trace IDs as hex strings
`SpanEventBridgeLogRecordProcessor` compares trace IDs as hex strings which forces the lazy hex encoding of four IDs per record. We should use `traceIdBytes` and equivalents instead.

Shaurya2k06:
## Goal
Match log records to the current span by comparing traceIdBytes and
spanIdBytes instead of hex strings, so emit does not force lazy hex
encoding of four IDs per record.

## Testing
Existing bridge/mismatch tests still pass. Added a case with copied
ID byte arrays so matching is by content, not array identity. Ran
span-event-bridge jvmTest plus detekt.

Fixes #969


fractalwrench (approved):
LGTM - thanks!

PR #983, opened 2026-08-25
fix: back in-memory exporters with threadSafeList

### LINKED ISSUE #966: In-memory exporters are backed by an unsynchronised ArrayList
In-memory exporters are backed by an unsynchronised ArrayList. `threadSafeList` or a lock would help prevent concurrent modification.

Shaurya2k06:
## Goal
Back in-memory span and log exporters with threadSafeList so concurrent
export/read does not hit an unsynchronised ArrayList.

## Testing
Existing InMemorySpanExporterTest and InMemoryLogRecordExporterTest cover
export, snapshots, and shutdown. Ran exporters-in-memory jvmTest plus detekt.

Fixes #966

### DIFF
--- a/exporters-in-memory/build.gradle.kts
+++ b/exporters-in-memory/build.gradle.kts
@@ -13,6 +13,7 @@ kotlin {
             dependencies {
                 implementation(project(":sdk-api"))
                 implementation(project(":sdk-common"))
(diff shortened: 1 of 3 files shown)


fractalwrench (approved):
LGTM - thanks!

PR #985, opened 2026-08-25
fix: apply W3C tracestate 512-character limit on decode

### LINKED ISSUE #965: Enforce W3CTraceContext limits on decode
Our W3CTraceContext implementation only enforces size limits when encoding, not on extract. We should alter the decode step so that it applies the limits correctly.

Shaurya2k06:
## Goal
Apply W3C tracestate combined-header limits (512 characters; drop members longer than 128 first, then from the end) on decode, not only via the 32-entry put cap. Encode uses the same truncation so inject and extract stay consistent. Invalid members are still skipped; a bad tracestate does not fail traceparent extract.

## Testing
Added TraceStateMarshaller tests for the 512/128 truncation on encode and decode, plus a W3CTraceContextPropagator extract case with an oversized tracestate. Ran implementation jvmTest for those classes plus detekt.

Fixes #965

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceStateMarshaller.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/propagation/TraceStateMarshaller.kt
@@ -4,6 +4,9 @@ import io.opentelemetry.kotlin.ExperimentalApi
 import io.opentelemetry.kotlin.factory.TraceStateFactory
 import io.opentelemetry.kotlin.tracing.TraceState
 
+internal const val MAX_TRACESTATE_CHARS = 512
+internal const val LARGE_MEMBER_CHARS = 128
+
 /**
  * Implementation of a W3C `tracestate` header.
  *
@@ -16,7 +19,7 @@ internal class TraceStateMarshaller(internal val traceState: TraceState) {
         traceState.asMap()
     }
 
-    fun encode(): String = W3CTraceStateCodec.encode(state)
+    fun encode(): String = W3CTraceStateCodec.encode(state.truncatedToHeaderLimit())
 
     companion object {
         fun decode(header: String, traceStateFactory: TraceStateFactory): TraceStateMarshaller {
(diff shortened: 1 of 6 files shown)


breedx-splk:
Thanks for jumping on this @Shaurya2k06. Frankly, I'm not convinced on this approach. I'm not sure that decoding the header into a TraceState map, only then to walk the map in order to reencode the map to a string to check the length -- seems kinda awkward/kludgey. 

In Java, we have chosen to simply limit the number of entries (not string/total length). See https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/propagation/internal/W3CTraceContextEncoding.java#L41C28-L41C44. That limits to 32 pairs, which has served us well and I think is a reasonable limit.

If the goal of #965 is to try and play nicely with the w3c spec, that's fine, but maybe we can reconsider the implementation? Curious what @fractalwrench thinks too.

fractalwrench:
I agree that creating an extra map isn't the optimal solution. If we're going to sanitize entries then it should probably happen within `W3CTraceStateCodec` (both for encoding and decoding).

The relevant requirements from the W3C TraceContext [spec are](https://www.w3.org/TR/trace-context/):

3.3.1.1 tracestate Header Field Values:
>There can be a maximum of 32 list-members in a list.

3.3.1.5 tracestate Limits:
>There are systems where propagating of 512 characters of tracestate may be expensive. In this case, the maximum size of the propagated tracestate header SHOULD be documented and explained. 

3.3.1.5 tracestate Limits:
>In a situation where tracestate needs to be truncated due to size limitations, the vendor MUST truncate whole entries. Entries larger than 128 characters long SHOULD be removed first. Then entries SHOULD be removed starting from the end of tracestate.

I would like to match all those requirements assuming that it's not massively onerous, as it enforces a limit on the entry length (whereas it sounds like opentelemetry-java's approach is technically unbounded)

Shaurya2k06:
Moved the 32/512/128 truncation into W3CTraceStateCodec for both encode and decode, i theres no more map round-trip, just member lengths 👍

fractalwrench (approved):
LGTM - thanks!

breedx-splk (approved):
Thanks!!

Would be nice to have a test that shows this interaction with TraceState validation. Specifically, it would be nice to have a test that shows a valid kv entry being dropped due to length truncation, even tho an invalid kv entry could have been retained through truncation only to be dropped during validation. It's a weird edge case, and shouldn't hold this up tho.

PR #988, opened 2026-08-26
fix: enhance scopeCacheKey to support attributes for deduplication

### LINKED ISSUE #930: Compat providers should use attributes in scope key
`LoggerProviderAdapter` doesn't use `attributes` as part of the scope key, whereas `LoggerProviderImpl` does. We should fix the compat implementation for logs/traces/metrics so their adapters use `attributes` as part of their scope key.

aaravmahajanofficial:
## Goal

`LoggerProviderAdapter` doesn't use attributes as part of the scope key, whereas `LoggerProviderImpl` does. Added fixed compat implementation for logs/traces/metrics so their adapters use attributes as part of their scope key. 

Fixes #930 

## Testing

- Compiled for all platforms. Detekt, apiCheck, and Android test compilation passed.


fractalwrench (approved):
LGTM - thanks!

PR #990, opened 2026-08-26
fix: enhance HttpClientRegistry to reuse default engine across exporter calls

### LINKED ISSUE #964: Use `HttpClientRegistry`
#789 added `HttpClientRegistry` but it's not currently used as there's a default argument that constructs a new engine per call:

```kotlin
public fun TraceExportConfigDsl.otlpHttpSpanExporter(
    baseUrl: String,
    httpClientEngine: HttpClientEngine = createHttpEngine(),   // new OkHttp/CIO engine every call
    timeoutMs: Long = EXPORT_REQUEST_TIMEOUT_MS,
)
```

aaravmahajanofficial:
## Goal

Fixes #964 (follow-up to #789).

Previously, exporter DSL entry points used `httpClientEngine: HttpClientEngine = createHttpEngine()` as a default parameter. Because default arguments are evaluated per call site in Kotlin, each call instantiated a fresh engine, causing `HttpClientRegistry` cache lookups to miss and repeatedly allocate new `HttpClient` instances.

This change:
- Introduces a lazily initialized `defaultEngine` inside `HttpClientRegistry` so unconfigured requests share a single engine instance.
- Updates `HttpClientRegistry.getOrCreate` to accept `engine: HttpClientEngine? = null` and fall back to `defaultEngine`.
- Updates `otlpHttpSpanExporter` and `otlpHttpLogRecordExporter` to default `httpClientEngine` to `null`.

## Testing

- Added `testDefaultPathReusesClientAcrossCalls` in `OtlpHttpSpanExporterTest`

### DIFF
--- a/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/HttpClientInstance.kt
+++ b/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/HttpClientInstance.kt
@@ -20,12 +20,11 @@ internal object HttpClientRegistry {
         clients.clear()
     }
 
-    fun getOrCreate(requestTimeoutMs: Long): HttpClient = getOrCreate(defaultEngine, requestTimeoutMs)
-
-    fun getOrCreate(engine: HttpClientEngine, requestTimeoutMs: Long): HttpClient {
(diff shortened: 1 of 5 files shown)


fractalwrench (approved):
LGTM - thanks!

PR #993, opened 2026-08-26
fix: call onEnding after releasing the span write lock

### LINKED ISSUE #970: User-supplied processor code runs while holding the span's write lock
`SpanModel.endInternal` invokes `processor.onEnding(ReadWriteSpanImpl(this))` inside `lock.write`,
and `addEvent`/`addLink` invoke the caller's `AttributesMutator` lambda inside `mutate { }`, which
is also the write lock.

Reads from the same thread are fine, but a user could plausibly hand the reference to another thread and block
waiting for it, which would deadlock. We should avoid the potential for this by ensuring `processor.onEnding` isn't called from within `lock.write`, but simultaneously we do need to enforce that `endInternal` can only run once.

Shaurya2k06:
## Goal
Call SpanProcessor.onEnding after releasing SpanModel’s write lock so a processor that hands the ReadWriteSpan to another thread cannot deadlock. endInternal still runs once via STARTED → ENDING → ENDED; onEnd stays outside the lock and still sees onEnding mutations.

## Testing
Existing span-end / onEnding tests plus a JVM test that mutates the span from another thread during onEnding. Ran implementation jvmTest and detekt.

Fixes #970

### DIFF
--- a/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt
+++ b/implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt
@@ -119,15 +119,21 @@ internal class SpanModel(
     @Suppress("NOTHING_TO_INLINE")
     private inline fun endInternal(timestamp: Long) {
         sdkErrorHandler.guard("Span.end failed") {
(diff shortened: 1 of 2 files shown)


fractalwrench (approved):
LGTM - thanks

Issue #974, opened 2026-08-24
OTLP exporter configuration surface


fractalwrench:
## EXP-3. 

`otlpHttpSpanExporter` / `otlpHttpLogRecordExporter` only accept
`(baseUrl, httpClientEngine, timeoutMs)` or `(baseUrl, httpClient)`. The spec however allows the following configuration options:

- **Headers.**
- **Compression.**
- **Endpoint semantics.** (the spec allows signal-agnostic endpoints, currently it's "$baseUrl/${endpoint.path}")
- **Certificates / mTLS.**
- **Protocol selection.**

**Scope:**

- [x] Add a DSL similar to `OpenTelemetryConfigDsl` that is only used as a function parameter on `otlpHttpSpanExporter` and equivalent
- [ ] Add options to the DSL to configure all items allowed in the spec

**Spec links:**

- OTLP exporter configuration: https://opentelemetry.io/docs/specs/otel/protocol/exporter/#configuration-options
- Env vars: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#otlp-exporter

Shaurya2k06:
scope seems pretty clear, I'm happy to take this up. 
cc @fractalwrench

Shaurya2k06:
I'm thinking of splitting this into multiple short PRS obv, first the DSL with endpoint/timeout/headers (existing overloads will stay). Then gzip, endpoint URL rules, TLS, and protocol later as we grow the transports.

Shaurya2k06:
Check out how I implemented the first one, I'll make the follow ups once given the green light.

PR #994, opened 2026-08-26
feat: add OTLP HTTP exporter DSL for endpoint, timeout, and headers

Shaurya2k06:
## Goal
Add an OtlpHttpExporterConfigDsl used as the lambda on otlpHttpSpanExporter / otlpHttpLogRecordExporter so endpoint, timeout, and headers can be set without a custom HttpClient. Existing overloads are unchanged. Gzip, endpoint-path rules, TLS, and protocol are follow-ups.

## Testing
OtlpClient and HTTP exporter tests for custom headers plus User-Agent. Ran exporters-otlp jvmTest, detekt, and apiDump.

Towards #974


exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpHttpExporterConfigDsl.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.opentelemetry.kotlin.init.ConfigDsl
+
+internal const val DEFAULT_OTLP_HTTP_ENDPOINT = "http://localhost:4318"
+
+/**
+ * Configures an OTLP HTTP exporter.
+ */
+@ExperimentalApi
+@ConfigDsl
+public class OtlpHttpExporterConfigDsl internal constructor() {
```
fractalwrench: Given this is a public API, could you make this an interface, then create an internal/private class that implements the interface? That follows patterns the rest of the project is already using.

exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/logging/export/OtlpLogRecordExporterApi.kt
```diff
@@ -9,10 +9,27 @@ import io.opentelemetry.kotlin.export.EXPORT_MAX_ATTEMPTS
...
+import io.opentelemetry.kotlin.export.OtlpHttpExporterConfigDsl
 import io.opentelemetry.kotlin.export.createDefaultHttpClient
 import io.opentelemetry.kotlin.export.createHttpEngine
+import io.opentelemetry.kotlin.export.createOtlpHttpClient
 import io.opentelemetry.kotlin.init.LogExportConfigDsl
 
+/**
+ * Creates a log record exporter that sends telemetry over OTLP HTTP.
+ */
+@ExperimentalApi
```
fractalwrench, line 14: Rather than adding another overload, I'd favour just removing the existing `otlpHttpLogRecordExporter` APIs (and similar for spans) as long as there is feature parity in `OtlpHttpExporterConfigDsl`. The function is marked with `ExperimentalApi` so it's reasonable for consumers to expect breaking changes. Same goes for the span variant of this function.

exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpHttpExporterConfigDsl.kt
```diff
@@ -0,0 +1,56 @@
...
+    public var timeoutMs: Long = EXPORT_REQUEST_TIMEOUT_MS
+
+    internal val headers: MutableMap<String, String> = linkedMapOf()
+    internal var httpClientEngine: HttpClientEngine? = null
+
+    /**
+     * Adds a header to every export request. A later call with the same [name] replaces the previous value.
+     * Blank names are ignored.
+     */
+    public fun header(name: String, value: String) {
```
fractalwrench: IMO let's skip this for now and just focus on the existing parameters that are supported when creating OTLP exporters. There are questions to answer around how to set headers properly, given that one use-case around this is setting Oauth tokens that aren't known up-front

fractalwrench (commented):
Thanks for starting this off. I left a few comments inline

fractalwrench (approved):
LGTM - thanks!

PR #996, opened 2026-08-27
Implement log and span processor behavior

RabahSend:
## Goal

Implements processor behavior on `LoggerProviderBehavior` and `TracerProviderBehavior`, and maps the existing DSL `export { }` onto it. This is not currently hooked up to the SDK but will be used to drive its behavior with resolved config.

## Testing
- [ ] `./gradlew :behavior:jvmTest :config-dsl:jvmTest`

Part of #924


behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/LogRecordProcessorBehavior.kt
```diff
@@ -0,0 +1,18 @@
...
+ * https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecordprocessor
+ */
+@ExperimentalApi
+class LogRecordProcessorBehavior : Behavior<LogRecordProcessorBehavior> {
+
+    override fun mergeWith(higher: LogRecordProcessorBehavior): LogRecordProcessorBehavior = higher
+
+    override fun equals(other: Any?): Boolean = other is LogRecordProcessorBehavior
+
+    override fun hashCode(): Int = 0
```
fractalwrench, line 17: Why are these specified manually rather than relying on a `data` class, and why is the hashCode always 0?

RabahSend: These types have no properties yet (presence only). A Kotlin data class needs at least one constructor property, so they can't be data classes as-is

equals treats every instance as the same configured marker. hashCode is 0 to stay consistent with that (all equal objects share one hash)

Once processor has fields (simple/batch, exporter), this should become a data class and these overrides go away

fractalwrench (approved):
LGTM - thanks!

PR #997, opened 2026-08-28
fix: enable gzip compression for OTLP HTTP requests

### LINKED ISSUE #977: OTLP HTTP exporter does not gzip-compress requests
### Describe the bug

The default OTLP HTTP client installs Ktor's `ContentEncoding` plugin with gzip enabled, and `OtlpClient` calls `compress("gzip")` for every request.

However, the resulting OTLP requests are not compressed:
```text
Content-Encoding: identity
encodedBytes=2219520
protobufBytes=2219520
```

Ktor's ContentEncoding plugin defaults to response decompression only. Request compression needs to be explicitly enabled using `ContentEncodingConfig.Mode.CompressRequest`.

This was discovered while investigating #796.

Related: #974 covers exposing compression as an exporter configuration option, while this issue concerns the current default client's behavior.

### Environment
  - Ktor: 3.2.3
  - JVM: 21
  - macOS

iliferov:
## Goal

  Fixes #977.

  The default OTLP HTTP client already registers gzip and calls `compress("gzip")`, but
  Ktor's `ContentEncoding` plugin defaults to response decompression only. This change
  configures it with `Mode.All`, enabling request compression while preserving response
  decompression.

  OTLP HTTP requests now include `Content-Encoding: gzip` and send gzip-compressed protobuf
  payloads.

  While updating the tests, two implementation details had to be handled:

  - Ktor produces the compressed request body lazily and ties it to the request execution
  context. Reading it later through `MockEngine.requestHistory` fails after that context
  completes, so tests now capture the body inside the mock handler.
  - Exporters run on `Dispatchers.Default` using real time, while `runTest` uses virtual
  time. Telemetry polling now runs on `Dispatchers.Default.limitedParallelism(1)` so its
  delays and timeouts use real time as well.

  The fake OTLP server used by smoke tests now decompresses request bodies before decoding
  protobuf telemetry.

  ## Testing

  - Added assertions for `Content-Encoding: gzip`.
  - Verified that decompressed request bodies match the original protobuf payloads.
  - Updated exporter and smoke tests to consume compressed request bodies.
  - Ran `./gradlew build -Psigning.skip=true`.

### DIFF
--- a/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/HttpClientInstance.kt
+++ b/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/HttpClientInstance.kt
@@ -4,6 +4,7 @@ import io.ktor.client.HttpClient
 import io.ktor.client.engine.HttpClientEngine
 import io.ktor.client.plugins.HttpTimeout
 import io.ktor.client.plugins.compression.ContentEncoding
+import io.ktor.client.plugins.compression.ContentEncodingConfig
 import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.util.collections.ConcurrentMap
 
@@ -38,6 +39,7 @@ internal fun createDefaultHttpClient(
     }
     install(ContentNegotiation)
     install(ContentEncoding) {
+        mode = ContentEncodingConfig.Mode.All
         gzip()
         deflate()
     }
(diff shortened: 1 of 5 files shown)


priettt (approved):
Thanks for contributing, looks good! 👍

breedx-splk (approved):
Thanks @iliferov ! 

I think long term, and within the scope of #974, we will need to allow users to disable the  compression, but that will be a future follow-up effort. Cheers! 🍻

fractalwrench (approved):
LGTM - thanks!

PR #998, opened 2026-08-29
docs: document snapshot publishing and Android SDK 36 requirements

iliferov:
## Goal

  Update the local development instructions in `CONTRIBUTING.md`:

  - Update the required Android SDK platform to 36 and Build Tools to 36.0.0.
  - Direct initial build verification to the full build instructions.
  - Add `-PsnapshotPublish=true` when publishing artifacts to Maven Local.

  The integration test fixtures consume `-SNAPSHOT` versions. Without `snapshotPublish`, `publishToMavenLocal`
  publishes release-versioned artifacts and the subsequent build cannot resolve the expected dependencies.

  ## Testing

  - Verified the Markdown link and formatting.
  - Ran `git diff --check`.
  - Documentation-only change; no build was run.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,8 +11,8 @@ would be accepted.
    1. JDK >=11 (OpenJDK 21 via https://sdkman.io/ is recommended)
    2. **Android SDK** — Android Studio bundles the SDK and is the easiest route: https://developer.android.com/studio.
       If installing the SDK manually, ensure the following packages are present:
-      - `platforms;android-34`
-      - `build-tools;30.0.3`
+      - `platforms;android-36`
+      - `build-tools;36.0.0`
       - `platform-tools`
    3. **Xcode** (macOS only) — the full Xcode app is required, not just the command-line tools.
       Install from the Mac App Store, then run:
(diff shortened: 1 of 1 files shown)


priettt (approved):

CONTRIBUTING.md
```diff
@@ -37,8 +38,8 @@ Maven Local first. This is a separate step because running both in the same Grad
 a race condition with Kotlin/Native:
 
 ```bash
-./gradlew publishToMavenLocal   # publishes snapshot artifacts to ~/.m2
-./gradlew build                  # runs all tests including the integration test
+./gradlew publishToMavenLocal -PsnapshotPublish=true   # publishes artifacts, including snapshot versions, to ~/.m2
```
breedx-splk: It would be ideal if this was implied and the user didn't have to do that, but thanks for capturing it.

breedx-splk: Maybe we can omit this part of the change and do https://github.com/open-telemetry/opentelemetry-kotlin/pull/1003 instead?

fractalwrench (approved):
LGTM - thanks!

PR #999, opened 2026-08-30
feat: Invoke shutdown/flush in compat mode

### LINKED ISSUE #792: Invoke shutdown/flush in compat mode
`CompatOpenTelemetryImpl` does not implement `TelemetryCloseable` so nothing is flushed or shutdown. We should attempt to invoke this.

Msksgm:
## Goal

Close https://github.com/open-telemetry/opentelemetry-kotlin/issues/792

Make lifecycle operations on the compat-mode `OpenTelemetry` implementation delegate to the underlying OpenTelemetry Java SDK tracer, logger, and meter providers.

The compat wrapper now:

- invokes all three providers for `forceFlush()` and `shutdown()` without short-circuiting;
- combines their results into a single success or failure result;
- applies an overall timeout;
- invokes provider shutdown only once; and
- treats Java API-only and no-op providers as successful no-op lifecycle operations.

## Testing

Added tests covering:

- successful `forceFlush()` and `shutdown()` delegation to all three Java SDK providers;
- failure aggregation while continuing to invoke the remaining providers;
- repeated shutdown without invoking provider shutdown more than once;
- the compat SDK construction path; and
- successful no-op lifecycle operations when wrapping the Java no-op API.

The `gradle-test`, `integration-test`, and required status checks pass.

### Coverage

Codecov reports 62.5% patch coverage, above the repository configured 50% patch target, and both `codecov/patch` and `codecov/project` pass.

The uncovered lines are defensive successful no-op branches for non-`TelemetryCloseable` providers. They are not reachable through the current public construction paths, which wrap providers in `TelemetryCloseable` adapters. The SDK-backed and Java API-only/no-op paths, delegation to all providers, success/failure aggregation, and idempotent shutdown behavior are covered.


priettt (approved):
looks good 👍

fractalwrench (approved):
LGTM

PR #1003, opened 2026-08-31
Always imply snapshot when publishing to maven local.

breedx-splk:
Related to #998.

When developers are building locally and want to publish to maven local for testing, lets make that easier thank requiring them to supply `-PsnpashotPublish=true`.


fractalwrench (approved):
LGTM

PR #1005, opened 2026-09-01
Alter JDK used in CI

fractalwrench:
## Goal

#1002 failed as the setup-java action has removed support for the 'adopt' distribution.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -26,9 +26,9 @@ jobs:
           persist-credentials: false
 
       - name: Setup Java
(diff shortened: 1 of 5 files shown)


bidetofevil (approved):
LGTM

CLOVIS-AI:
Note that you can use [Gradle Daemon Toolchains](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:configuring_daemon_jvm) for Gradle to auto-download the JVM you want. This guarantees that every developer gets the exact same JVM on their machine as well.

PR #1006, opened 2026-09-01
Bump compose dependency

fractalwrench:
## Goal

Bumps the compose dependency. This required bumping the compileSdk as #1000 failed otherwise (this should have no effect on library consumers).


bidetofevil (approved):
LGTM

PR #1007, opened 2026-09-01
Alter stale labelling workflows

fractalwrench:
## Goal

Alters the github workflows that mark PRs as needing author feedback/stale so that they have the correct permissions to write labels. Additionally I've altered the workflows so that they respond when comments are added as replies on a PR.

I've decreased the stale limit to 14 days as this gives 28 days in total after feedback was requested before the PR auto-closes, which feels reasonable given the current pace of the project.


bidetofevil (approved):
LGTM. May want a maintainer to approve this change too though, explicitly via review or verbally.

PR #1008, opened 2026-09-01
Refactor SDK to use attribute limits supplied via AttributeLimitsBehavior

fractalwrench:
## Goal

Refactors the SDK so that compat/implementation use `AttributeLimitsBehavior`. This is the next step in hooking up the SDK to use the config resolved from YAML/envar/DSL sources.

## Testing

Updated existing unit tests.


behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/LogLimitsBehavior.kt
```diff
@@ -3,26 +3,10 @@ package io.opentelemetry.kotlin.behavior
...
-     */
-    val attributeValueLengthLimit: Int? = null,
-) : Behavior<LogLimitsBehavior> {
-
-    override fun mergeWith(higher: LogLimitsBehavior): LogLimitsBehavior = copy(
-        attributeCountLimit = higher.attributeCountLimit ?: attributeCountLimit,
-        attributeValueLengthLimit = higher.attributeValueLengthLimit ?: attributeValueLengthLimit,
-    )
-}
+typealias LogLimitsBehavior = AttributeLimitsBehavior
```
bidetofevil, line 12: Do we really need this alias or can we just reference the actual underlying class?

bidetofevil (approved):
LGTM. One non-blocking comment about getting rid of the alias if there's no good reason to maintain it. If it blows up the PR size, maybe we can do it in a subsequent PR?

behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/LogLimitsBehavior.kt
fractalwrench: IMO it's helpful as `AttributeLimitsBehavior` could be confused with the global attribute limits. I also assume that additional log limits may be added in future..

PR #1009, opened 2026-09-01
Create envar reader for config

fractalwrench:
## Goal

Creates a class that reads all the envars that config currently supports and builds an `OpenTelemetryBehavior` from it.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #1013, opened 2026-09-03
Read config yaml on SDK init

fractalwrench:
## Goal

Hooks up SDK init so that it reads and resolves config across all its sources (DSL/YAML/envars). A future changeset will hook up the `OpenTelemetryBehavior` to the actual SDK in implementation and compat modules.


bidetofevil (approved):
LGTM.

Perhaps this is something we should look at when we hook this up: are we already paying the penalty to read a file on startup (in a blocking way) and is there at least a path such that we don't pay the costs if there are no config YAMLs to process?

fractalwrench:
The YAML config read is only attempted on the JVM target (not Android) for now. When we support Android, we'll need to figure out an alternative mechanism of reading the YAML IMO.

Issue #906, opened 2026-08-21
Configuration: SpanLimits


fractalwrench:
## SpanLimits

SpanLimits: enforces limits on span telemetry.
Contains the following fields: `attribute_count_limit`, `attribute_value_length_limit`,
`event_count_limit`, `link_count_limit`, `event_attribute_count_limit`, `link_attribute_count_limit`

Declarative config path: `.tracer_provider.limits`
Schema type: `SpanLimits`
Env vars: `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`, `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT`,
`OTEL_SPAN_EVENT_COUNT_LIMIT`, `OTEL_SPAN_LINK_COUNT_LIMIT`, `OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT`,
`OTEL_LINK_ATTRIBUTE_COUNT_LIMIT`

Spec links:

- Behavior: https://opentelemetry.io/docs/specs/otel/trace/sdk/#span-limits
- Shared attribute semantics: https://opentelemetry.io/docs/specs/otel/common/#attribute-limits
- Env vars: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#span-limits
- YAML schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/tracer_provider.yaml

Implementation checklist:

- [x] Behavior implemented in `behavior` module? — `SpanLimitsBehavior`
- [x] Readable from `config-dsl` module? — `SpanLimitsConfigDslImpl`
- [x] Readable from `config-envar` module? — `SpanLimitsEnvVars`
- [x] Readable from `config-yaml` module? — `SpanLimitsMapper`
- [x] Compat mode SDK obeys config
- [x] Implementation mode SDK obeys config

PR #1017, opened 2026-09-07
feat: add implementation mode sdk obeys config

Msksgm:
## Goal

Route span limits configured through the DSL through `SpanLimitsBehavior` in the implementation SDK.

Include the tracer provider's span limits in behavior resolution, then convert the resolved values into `SpanLimitConfig`. Preserve fallback to global attribute limits and SDK defaults, and remove the unused `SpanLimitsConfigImpl`.

This covers the implementation SDK's DSL path. Compat mode and wiring YAML/environment variables into SDK initialization remain separate work.

Refs #906.

## Testing

- Verify all six span limits are mapped from `SpanLimitsBehavior` into `SpanLimitConfig`.
- Verify DSL-configured limits reach the generated tracing configuration.
- Existing tests cover defaults, global attribute limit fallback, and zero-value precedence.


fractalwrench (approved):
LGTM - thanks!

PR #1022, opened 2026-09-07
Create context-impl module to separate out context/propagator internals

fractalwrench:
## Goal

This breaks a smaller changeset off the POC in #979 that aims to make the propagator implementation part of the official API. This changeset is an internal refactor that adds a non-public module named `context-impl`, and effectively holds internal implementation details such as Baggage implementation in a module that could be accessed from both compat/implementation.

## Testing


bidetofevil (approved):
LGTM

PR #1026, opened 2026-09-09
Alter DSL sourceSet syntax

fractalwrench:
## Goal

Updates the syntax used for configuring sourceSets, as the previous mechanism has been deprecated since Gradle 9.6.


bidetofevil (approved):
LGTM

PR #1027, opened 2026-09-09
Move propagators to context-impl module

fractalwrench:
## Goal

Builds on #1022 by moving more propagators to the context-impl module. These will eventually allow the propagators to be created independent of the OpenTelemetry API.

## Testing


bidetofevil (approved):
Nice. I guess I never realized how not tethered to OTel APIs this API actually is.

PR #1029, opened 2026-09-09
feat: map console exporter YAML and env vars to processor behavior

RabahSend:
## Goal

Map ConsoleExporter from declarative YAML and environment variables onto `processor.console` on tracer and logger providers. Part of #924.

YAML: if any simple or batch processor has `exporter.console`, set `SpanProcessorBehavior` / `LogRecordProcessorBehavior` with `ConsoleExporterBehavior`. Other processors/exporters are left unset.

Env: `OTEL_TRACES_EXPORTER=console` and `OTEL_LOGS_EXPORTER=console` do the same. Known values that this MVP does not map (`otlp`, `zipkin`, `none`, `logging`, `otlp/stdout`) leave the processor unset. Unknown values are ignored and reported via `onWarning`.

This does not make `implementation` or `compat` obey the resolved behavior.

## Testing

`./gradlew :config-yaml:jvmTest :config-envar:jvmTest`

### DIFF
--- a/config-envar/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/tracing/TracesExporterEnvVars.kt
+++ b/config-envar/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/tracing/TracesExporterEnvVars.kt
@@ -20,7 +20,7 @@ class TracesExporterEnvVars(
         val name = reader.readString(EXPORTER)?.takeIf { it.isNotEmpty() } ?: return null
         return when (name.lowercase()) {
             CONSOLE -> SpanProcessorBehavior(console = ConsoleExporterBehavior())
-            OTLP, ZIPKIN, LOGGING, NONE, OTLP_STDOUT -> null
+            OTLP, LOGGING, NONE, OTLP_STDOUT -> null
             else -> null.also { onWarning("Unknown OTEL_TRACES_EXPORTER value '$name'; ignoring") }
         }
     }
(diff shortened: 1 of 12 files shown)


config-envar/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/tracing/TracesExporterEnvVars.kt
```diff
@@ -0,0 +1,37 @@
...
+            OTLP, ZIPKIN, LOGGING, NONE, OTLP_STDOUT -> null
+            else -> null.also { onWarning("Unknown OTEL_TRACES_EXPORTER value '$name'; ignoring") }
+        }
+    }
+
+    private companion object {
+        const val EXPORTER = "OTEL_TRACES_EXPORTER"
+        const val CONSOLE = "console"
+        const val OTLP = "otlp"
+        const val ZIPKIN = "zipkin"
```
fractalwrench: zipkin exporter is deprecated and will be removed by the end of the year - IMO we should just remove it as an option in this PR and don't need to support it.

fractalwrench (commented):
One comment, otherwise this is good to go

config-envar/src/commonMain/kotlin/io/opentelemetry/kotlin/config/envar/tracing/TracesExporterEnvVars.kt
RabahSend: Got it, I’ll do that.

fractalwrench (approved):
LGTM - thanks!

PR #1030, opened 2026-09-09
feat: add compat mode sdk obeys config

Msksgm:
## Goal

Route span limits through `SpanLimitsBehavior` in the compat SDK, following the implementation SDK changes in #1017.

Include the tracer provider's DSL span limits in behavior resolution, then apply the resolved values to the Java SDK and compat adapters. Preserve fallback to global attribute limits and existing defaults.

Refs https://github.com/open-telemetry/opentelemetry-kotlin/issues/906

## Testing

- Verify all six DSL-configured span limits reach the resolved behavior.
- Update existing tests to pass `SpanLimitsBehavior` explicitly, preserving coverage for defaults, global attribute limit fallback, partial overrides, and zero-value precedence.


fractalwrench (approved):
LGTM - thanks!

PR #1034, opened 2026-09-11
Allow supplying clock param

fractalwrench:
## Goal

When calling `toOtelKotlinApi` it should be possible to pass in the `Clock` instance if this is known, rather than assuming the default is used. This is not the case for opentelemetry-android, for example, which uses a custom implementation, and we have no way to obtain it from this project so require users to specify a parameter if they care about this.

We still supply the default clock instance if nothing is supplied.


bidetofevil (approved):
LGTM

Issue #908, opened 2026-08-21
Configuration: LogRecordLimits


fractalwrench:
## LogRecordLimits

LogRecordLimits: enforces limits on log record telemetry.
Contains the following fields: `attribute_count_limit`, `attribute_value_length_limit`

Declarative config path: `.logger_provider.limits`
Schema type: `LogRecordLimits`
Env vars: `OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT`, `OTEL_LOGRECORD_ATTRIBUTE_VALUE_LENGTH_LIMIT`

Spec links:

- Behavior: https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecord-limits
- Shared attribute semantics: https://opentelemetry.io/docs/specs/otel/common/#attribute-limits
- Env vars: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#logrecord-limits
- YAML schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/logger_provider.yaml

Implementation checklist:

- [x] Behavior implemented in `behavior` module? — `LogLimitsBehavior`
- [x] Readable from `config-dsl` module? — `LogLimitsConfigDslImpl`
- [x] Readable from `config-envar` module? — `LogLimitsEnvVars`
- [x] Readable from `config-yaml` module? — `LogRecordLimitsMapper`
- [x] Compat mode SDK obeys config
- [x] Implementation mode SDK obeys config
- [x] Overrides the global `.attribute_limits` equivalents

PR #1036, opened 2026-09-13
feat: Configuration LogRecordLimits add implementation mode sdk obeys config

Msksgm:
## Goal

Route log record limits through `LogLimitsBehavior` in the implementation SDK.

Include the logger provider's DSL limits in behavior resolution, then pass the resolved limits into `LoggingConfig`. Preserve per-field fallback to global attribute limits and existing SDK defaults.

Refs #908. Compat mode remains separate work.

## Testing

- Verify both log limits are applied from `LogLimitsBehavior`.
- Verify DSL-configured log limits take precedence over global attribute limits.
- Verify partially specified log limits inherit unset fields from global attribute limits.
- Preserve existing coverage for unspecified limits and zero-value precedence.
- Local JVM tests and Detekt checks pass.
- CI `gradle-test`, `integration-test`, and required status checks pass.


fractalwrench (approved):
LGTM - thanks!

PR #1042, opened 2026-09-14
feat: Configuration LogRecordLimits add compat mode sdk obeys config

Msksgm:
## Goal

Route log record limits through `LogLimitsBehavior` in compat mode, following the span limits changes in #1030.

Include the logger provider's DSL limits in behavior resolution, then pass the resolved limits to the Java SDK. Preserve per-field fallback to global attribute limits and the Java SDK's defaults.

Refs #908. Complements the implementation mode changes in #1036.

## Testing

- Verify both DSL-configured log limits are reflected in the resolved behavior.
- Update existing tests to pass resolved log limits, preserving coverage for signal-specific precedence, zero values, and unset limits.
- Add coverage for partially specified log limits inheriting unset fields from global attribute limits.
- Existing log export coverage verifies attribute count and value length limits on emitted logs.
- Local JVM tests and Detekt checks pass.


fractalwrench (approved):
LGTM - thanks!

PR #1010, opened 2026-09-02
feat: add SamplerBehavior 

aaravmahajanofficial:
## Goal

Partially addresses #913

Implements the sampler configuration models within the `behavior` module

## Testing

- Added unit tests in `SamplerBehaviorTest` 
- Added unit tests in `TracerProviderBehaviorTest`

---

The work has been divided into multiple-PRs, linked below. To be reviewed and merged in this order:

- https://github.com/aaravmahajanofficial/opentelemetry-kotlin/pull/1
- https://github.com/aaravmahajanofficial/opentelemetry-kotlin/pull/2
- https://github.com/aaravmahajanofficial/opentelemetry-kotlin/pull/3
- https://github.com/aaravmahajanofficial/opentelemetry-kotlin/pull/5
- https://github.com/aaravmahajanofficial/opentelemetry-kotlin/pull/6
- https://github.com/aaravmahajanofficial/opentelemetry-kotlin/pull/7

---

## Pending Work
- Wiring Experimental Schemas: `composite/development`, `jaeger_remote/development`, `probability/development`

Issue #913, opened 2026-08-21
Configuration: Sampler


fractalwrench:
## Sampler

Sampler: chooses which sampler the TracerProvider uses.
Contains the following fields: `always_on`, `always_off`, `trace_id_ratio_based`, `parent_based`,
`composite/development`, `jaeger_remote/development`, `probability/development`

Declarative config path: `.tracer_provider.sampler`
Schema type: `Sampler`, `AlwaysOnSampler`, `AlwaysOffSampler`
Env vars: `OTEL_TRACES_SAMPLER`, `OTEL_TRACES_SAMPLER_ARG`

Spec links:

- Sampling: https://opentelemetry.io/docs/specs/otel/trace/sdk/#sampling
- Sampler interface: https://opentelemetry.io/docs/specs/otel/trace/sdk/#sampler
- Built-in samplers: https://opentelemetry.io/docs/specs/otel/trace/sdk/#built-in-samplers
- AlwaysOn / AlwaysOff: https://opentelemetry.io/docs/specs/otel/trace/sdk/#alwayson
- Env vars: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration
- YAML schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/tracer_provider.yaml

Implementation checklist:

- [ ] Behavior implemented in `behavior` module?
- [ ] Readable from `config-dsl` module?
- [ ] Readable from `config-envar` module?
- [ ] Readable from `config-yaml` module?
- [ ] Compat mode SDK obeys config
- [ ] Implementation mode SDK obeys config

aaravmahajanofficial:
Picked this.

PR #1043, opened 2026-09-14
feat: Kotlin SDK obey sampler configuration

aaravmahajanofficial:
Part of #913
Related to #1010 

## Goal

Wire the Kotlin implementation SDK to apply resolved sampler configuration.

## Testing

Add tests in:
 - `OpenTelemetryEnvVarsTest‎.kt`
- `OpenTelemetryConfigReaderTest.kt`
- `ResolvedSamplerConfigTest.kt`


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/init/OpenTelemetryConfigImpl.kt
```diff
@@ -39,6 +46,10 @@ internal class OpenTelemetryConfigImpl(
     internal val propagatorCfg: PropagatorConfigImpl = PropagatorConfigImpl()
     private val globalAttributeLimits = AttributeLimitsConfigDslImpl()
     private val resourceDetectionConfig = ResourceDetectionConfigImpl()
+    private val behaviorReader: BehaviorReader = suppliedBehaviorReader ?: defaultBehaviorReader(
+        envVarReader = envVarReader,
+        onSamplerWarning = ::reportSamplerWarning,
+    )
```
fractalwrench: `onSamplerWarning` feels like the wrong abstraction here as `defaultBehaviorReader` will grow to have a lot of parameters if a separate lambda is required for each potential warning. Passing `SdkErrorHandler` would probably be a better option here IMO.

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/init/TracerProviderConfigImpl.kt
```diff
@@ -32,6 +34,7 @@ internal class TracerProviderConfigImpl(
 
     private var processor: SpanProcessor? = null
     private var samplerAction: SamplerConfigDsl.() -> Sampler = { parentBased(root = alwaysOn()) }
+    private var samplerConfiguredByDsl = false
```
fractalwrench: This extra property could be removed and `samplerAction` made `null`, then `behaviorReader` higher up could be responsible for determining what the default sampler is.

aaravmahajanofficial: Thanks, resolved!

implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/init/OpenTelemetryConfigImpl.kt
aaravmahajanofficial: Resolved in new commit, thanks!

fractalwrench (approved):
LGTM - thanks!

PR #1047, opened 2026-09-15
Avoid calling setNoParent in TracerAdapter

fractalwrench:
## Goal

Avoid calling `setNoParent` in TracerAdapter as it incorrectly sets the span's context as the root context, whereas we want opentelemetry-java's current context.

## Testing

Added test coverage.


bidetofevil (approved):
Just to confirm: by default, if no current context is set, we will still get the same behaviour?

fractalwrench:
>Just to confirm: by default, if no current context is set, we will still get the same behaviour?

Previously for a child span the context was erroneously set as the root context, rather than the implicit context. This change fixes that. Log records were unaffected. We're now getting the same behavior as opentelemetry-java.

PR #1051, opened 2026-09-15
Handle attributes and links in compat sampling

fractalwrench:
## Goal

It's possible to specify attributes/links at span creation but these weren't being passed to the sampler in the compat module, whereas the spec states they should be visible to inform sampling decisions.

## Testing

Added test coverage.


bidetofevil (approved):
LGTM - though perhaps calling it something other than a `Collector` is a good idea given we already have Collectors in OTel?

fractalwrench:
Renamed class

PR #1052, opened 2026-09-15
Move symbols to context-impl module

fractalwrench:
## Goal

Move additional symbols to the `context-impl` module, which is a pre-requisite for creating a standalone Propagators API.


bidetofevil (approved):
LGTM. Interesting that this didn't surface earlier.

PR #1053, opened 2026-09-15
Simplify how global/local attribute limits are configured

fractalwrench:
## Goal

Simplifies how global/local attribute limits are configured for logs and spans by resolving precedence on the `OpenTelemetryBehavior` object for both implementation and compat.

## Testing

Updated unit tests.


bidetofevil (approved):
LGTM

PR #1054, opened 2026-09-16
Move the remainder of the propagation package to context-impl

fractalwrench:
## Goal

Moves the remainder of the propagation package to the context-impl module.


bidetofevil (approved):
LGTM

PR #1055, opened 2026-09-16
Avoid cast in baggage conversion

fractalwrench:
## Goal

Avoids a cast in baggage conversion that could potentially fail if `Baggage` is a different type.

## Testing

Added unit tests.


compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/baggage/BaggageExt.kt
```diff
@@ -0,0 +1,16 @@
+package io.opentelemetry.kotlin.baggage
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.aliases.OtelJavaBaggage
+import io.opentelemetry.kotlin.aliases.OtelJavaBaggageEntryMetadata
+
+@OptIn(ExperimentalApi::class)
+public fun Baggage.toOtelJavaBaggage(): OtelJavaBaggage {
+    (this as? BaggageAdapter)?.let { return it.impl }
```
bidetofevil: Purely an aesthetic thing but this line reads real ugly to me....

bidetofevil (approved):
LGTM. That one line to do an early return isn't my favourite, but that was probably chosen because the alternative is worse eh?

PR #1061, opened 2026-09-16
feat: Compat SDK obey sampler configuration

aaravmahajanofficial:
Part of #913
Related to #1010 

## Goal

Wire the Compat SDK to apply resolved sampler configuration.

## Testing

Add tests in:
- `CompatResolvedSamplerConfigTest.kt`
- `CompatSamplerBehaviorMappingTest.kt`


compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/CompatSamplerBehaviorMapping.kt
```diff
@@ -0,0 +1,19 @@
+package io.opentelemetry.kotlin.tracing.sampling
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.behavior.SamplerBehavior
+import io.opentelemetry.kotlin.init.SamplerConfigDsl
+
+@ExperimentalApi
+internal fun SamplerConfigDsl.toSampler(behavior: SamplerBehavior): Sampler = when (behavior) {
```
fractalwrench: This duplicates the same code in `implementation`. It'd be ideal if it could move to config-dsl (if possible)

compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/init/CompatTracerProviderConfig.kt
```diff
@@ -74,10 +77,21 @@ internal class CompatTracerProviderConfig(
...
+
+    internal fun applyResolvedSampler(behavior: SamplerBehavior?) {
+        if (sampler != null || behavior == null) {
+            return
         }
-        val sampler = samplerConfig.action()
+        sampler = newSamplerDsl().toSampler(behavior)
+    }
+
+    private fun newSamplerDsl(): SamplerConfigDsl = object : SamplerConfigDsl {
```
fractalwrench:
```suggestion
    private val newSamplerDsl: SamplerConfigDsl = object : SamplerConfigDsl {
```

compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/init/CompatResolvedSamplerConfigTest.kt
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    private fun env(sampler: String): (String) -> String? {
+        val values = buildMap {
+            put("OTEL_TRACES_SAMPLER", sampler)
+        }
+        return values::get
+    }
+
+    private fun writeConfigFile(contents: String): String {
```
fractalwrench: This doesn't seem great needing to write a file each time YAML is required. I think it'd be good to either rethink how this is being unit tested so that it doesn't require I/O, or design a way to pass in yaml without a config file (which feels out-of-scope)

compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/CompatSamplerBehaviorMapping.kt
aaravmahajanofficial: Moved the shared `when` / ParentBased defaults into `config-dsl` as `SamplerBehavior.mapToSampler`. The remaining SDK helpers only bind that to each SDK's `alwaysOn` / `alwaysOff` / `parentBased` factories — those still have to live in `implementation` vs `compat` because they construct different `Sampler` types.

compat/src/jvmTest/kotlin/io/opentelemetry/kotlin/init/CompatResolvedSamplerConfigTest.kt
aaravmahajanofficial: Dropped temp YAML files from this test. File parse/precedence is already covered in `config-yaml` and `OpenTelemetryConfigReaderTest`. This class now injects resolved `SamplerBehavior` when it needs to assert the Compat SDK applies that layer.

compat/src/jvmAndAndroidMain/kotlin/io/opentelemetry/kotlin/tracing/sampling/CompatSamplerBehaviorMapping.kt
aaravmahajanofficial: addressed this in new commit

fractalwrench (approved):
LGTM - thanks!

PR #1064, opened 2026-09-16
Fix concurrent span mutation during OnEnding

### LINKED ISSUE #1059: A span remains mutable from other threads while `OnEnding` runs
[The spec](https://opentelemetry.io/docs/specs/otel/trace/sdk/#onending) states
>"The SDK MUST guarantee that the
span can no longer be modified by any other thread before invoking `OnEnding` of the first `SpanProcessor`.
From that point on, modifications are only allowed synchronously from within the invoked `OnEnding`
callbacks."

Currently in `SpanModel` `endInternal` calls `onEnding` outside the lock, and as `mutate {}` gates on `isRecordingInternal()` it's possible to mutate the span concurrently.

0utsights:
## Goal

During `OnEnding`, another thread could overwrite span data before export. Restrict mutations in the `ENDING` state to the synchronous callback thread using the existing multiplatform `ThreadLocal`.

The thread-local is initialized lazily and used only when a processor requires `OnEnding`. Mutation checks do not initialize it. Callbacks remain outside the write lock to avoid deadlocks, and `finally` restores the callback marker.

## Testing

Validation base: `959ae4a8fea2275d0b53dd85a15634e5c907f96e` plus the review revision.

- The original regression reproduced cross-thread overwrites before the fix.
- All five `SpanOnEndingOtherThreadTest` tests passed, including the disabled-callback path, active-span writes, callback writes, rejected outside writes with worker completion, reentrant/repeated ending, immutable snapshots, and a throwing callback. Independent review reran these tests on the final revision.
- `:implementation:jvmTest` passed: 563 tests, zero failures/errors/skips.
- `:implementation:jsNodeTest` passed: 548 tests, zero failures/errors/skips.
- Repository-wide `detekt`, `:implementation:detektMetadataCommonMain`, and `:implementation:detektJvmTest` passed.
- `git diff --check` passed.

Local validation used Windows, JDK 21 to run Gradle, and JDK 11/17 compiler toolchains. Android and iOS checks were not run locally for this revision; they remain for upstream CI.

Fixes #1059.


implementation/src/commonMain/kotlin/io/opentelemetry/kotlin/tracing/model/SpanModel.kt
```diff
@@ -130,9 +134,15 @@ internal class SpanModel(
...
+            val previousEndingSpan = endingSpan.get()
+            endingSpan.set(this)
+            try {
+                sdkErrorHandler.guard {
+                    processor?.takeIf(SpanProcessor::isOnEndingRequired)
+                        ?.onEnding(ReadWriteSpanImpl(this))
+                }
+            } finally {
+                endingSpan.set(previousEndingSpan)
             }
```
fractalwrench: I think it'd be worth skipping the thread-local entirely if `isOnEndingRequired` is false. Making the property lazy would also help

fractalwrench (commented):
One comment, otherwise this looks good to go.

fractalwrench (approved):
LGTM - thanks!

Issue #905, opened 2026-08-21
Configuration: Resource


fractalwrench:
## Resource

Resource: the immutable set of attributes describing the entity producing telemetry.
Contains the following fields: `attributes` (list of name/value/type), `attributes_list`, `schema_url`

Declarative config path: `.resource`
Schema type: `Resource`
Env vars: `OTEL_RESOURCE_ATTRIBUTES`, `OTEL_SERVICE_NAME`, `OTEL_ENTITIES`

Spec links:

- Behavior: https://opentelemetry.io/docs/specs/otel/resource/sdk/
- Via env var: https://opentelemetry.io/docs/specs/otel/resource/sdk/#specifying-resource-information-via-an-environment-variable
- Env vars: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration
- YAML schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/resource.yaml

Implementation checklist:

- [ ] Behavior implemented in `behavior` module?
- [ ] Readable from `config-dsl` module?
- [ ] Readable from `config-envar` module?
- [ ] Readable from `config-yaml` module?
- [ ] Compat mode SDK obeys config
- [ ] Implementation mode SDK obeys config
- [ ] `OTEL_SERVICE_NAME` takes precedence over a `service.name` set via `OTEL_RESOURCE_ATTRIBUTES`.
- [ ] `OTEL_RESOURCE_ATTRIBUTES` uses W3C Baggage decoding

rastiehaiev:
Hi @Karthik-Chowdary, I saw that #1069 completed the `config-dsl` part of this issue. Are you planning to continue with the remaining parts? If not, I’d be interested in taking them.

PR #1069, opened 2026-09-17
feat: map resource DSL to behavior

Karthik-Chowdary:
## Summary

Implements the programmatic resource configuration mapping for the `config-dsl` module.

- maps `serviceName`, schema URL, and resource attributes to `ResourceBehavior`
- preserves the existing attribute type conversion behavior for map input
- defensively copies mutable attribute values, including nested `AnyValue` values
- adds JVM/JS-portable coverage for defaults, mapping, copies, and service-name precedence

This completes the **Readable from `config-dsl` module** part of #905.

## Testing

- `./gradlew :config-dsl:jvmTest --tests '*ResourceConfigDslImplTest'`
- `./gradlew :config-dsl:allTests`
- `./gradlew :config-dsl:detekt --no-configuration-cache`
- `git diff --check`

Fixes part of #905.


fractalwrench:
If you're able to sign the CLA someone will take a look at this. Thanks

Karthik-Chowdary:
The EasyCLA check is now passing for the current commit. Thank you — the PR is ready for review when convenient.

config-dsl/src/commonMain/kotlin/io/opentelemetry/kotlin/config/dsl/ResourceConfigDslImpl.kt
```diff
@@ -0,0 +1,109 @@
...
+        schemaUrl = schemaUrl,
+        attributes = attributes.toMap().takeUnless { it.isEmpty() },
+    )
+}
+
+private class ResourceAttributesMutator : AttributesMutator {
+    private val attributes = mutableMapOf<String, Any>()
+
+    fun putAll(values: Map<String, Any>) {
+        values.forEach { (key, value) ->
```
fractalwrench: If you add a dependency on the `api-ext` module then you can use `setAttributes(attributes: Map<String, Any>)` which achieves this under the hood.

fractalwrench (commented):
One comment about an API in the `api-ext` module that would simplify this PR quite a bit. The remainder of the PR looks good

config-dsl/src/commonMain/kotlin/io/opentelemetry/kotlin/config/dsl/ResourceConfigDslImpl.kt
Karthik-Chowdary: Addressed in c0ce95e3: `config-dsl` now depends on `api-ext` and delegates map conversion to `AttributesMutator.setAttributes`, removing the duplicate conversion logic.

fractalwrench (approved):
LGTM - thanks!

PR #1070, opened 2026-09-18
Reduce internal uses of `IdGenerator` interface for invalid IDs

fractalwrench:
## Goal

Reduces internal uses of the `IdGenerator` interface to get invalid IDs, as these can be supplied via constants instead.

## Testing


bidetofevil (approved):
LGTM

PR #1050, opened 2026-09-15
feat: IdGenerator config [4/4] - compatibility SDK obey ID generator configuration

rastiehaiev:
## Goal

Partially addresses #907 (4/4).

Makes the compatibility SDK honor the resolved ID generator behavior by mapping `IdGeneratorBehavior.Random` to `CompatIdGenerator`.

The resolution order remains:

1. A custom generator supplied through the DSL.
2. The resolved ID generator behavior.
3. The compatibility SDK's default `CompatIdGenerator`.

This is the final part of the ID generator configuration work.

## Testing

- Added unit coverage for mapping `IdGeneratorBehavior.Random`.
- Added coverage for the default generator, resolved behavior, and custom DSL precedence.
- Ran `./gradlew :compat:jvmTest detekt apiCheck`.
- Ran `./gradlew assembleAndroidTest`.

Issue #914, opened 2026-08-21
Configuration: ParentBasedSampler


fractalwrench:
## ParentBasedSampler

ParentBasedSampler: delegates the sampling decision based on the parent span context.
Contains the following fields: `root`, `remote_parent_sampled`, `remote_parent_not_sampled`,
`local_parent_sampled`, `local_parent_not_sampled`

Declarative config path: `.tracer_provider.sampler.parent_based`
Schema type: `ParentBasedSampler`
Env vars: `OTEL_TRACES_SAMPLER=parentbased_always_on` / `parentbased_always_off` /
`parentbased_traceidratio` / `parentbased_jaeger_remote`

Spec links:

- Behavior: https://opentelemetry.io/docs/specs/otel/trace/sdk/#parentbased
- Recording/Sampled reaction table: https://opentelemetry.io/docs/specs/otel/trace/sdk/#recording-sampled-reaction-table
- Env vars: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration
- YAML schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/tracer_provider.yaml

Implementation checklist:

- [ ] Behavior implemented in `behavior` module?
- [ ] Readable from `config-dsl` module?
- [ ] Readable from `config-envar` module?
- [ ] Readable from `config-yaml` module?
- [ ] Compat mode SDK obeys config
- [ ] Implementation mode SDK obeys config
- [ ] Cap nesting depth when parsing YAML to avoid stackoverflow

Suhail98:
Hi @fractalwrench — I'd like to take this one if it's free.

Reading through `main`, most of the checklist looks done already:

- `SamplerBehavior.ParentBased` in the `behavior` module
- `SamplerConfigDslImpl.parentBased(...)` for the DSL
- `SamplerMapper` for `tracer_provider.sampler.parent_based`
- `SamplerEnvVars` for `parentbased_always_on` / `parentbased_always_off`
- `SamplerBehaviorMapping`, so the Kotlin SDK obeys it

Two things still look open to me.

[...]

PR #1071, opened 2026-09-18
fix: cap parent_based sampler nesting when mapping YAML config

Suhail98:
## Goal

Partially addresses #914.

### The problem

In the generated schema a `parent_based` sampler holds five child samplers, and each of those can be another `parent_based` sampler. Nothing stops a config file from going round that loop forever:

```yaml
tracer_provider:
  sampler:
    parent_based:
      root:
        parent_based:
          root:
            parent_based: # ...and so on
```

`SamplerMapper.toBehavior()` followed that nesting with no limit, so a file like the one above threw `StackOverflowError` while the SDK was starting up — and an error thrown during init takes the host app with it.

### The fix

Check how deep the nesting goes before mapping anything, and stop at 10 levels.

- A file nesting 10 levels or fewer maps exactly as it did before.
- Past 10 levels the sampler is left unset as a whole, so the SDK falls back to its default rather than to a half-mapped version of what the file asked for. This is how the mapper already treats a file that configures two conflicting samplers.
- The depth check walks no more than 10 levels itself, so it does not recurse down the file it is rejecting.

The depth is checked up front rather than counted while mapping because a child left unset already means "use the spec default". If the mapper gave up partway down, there would be no way to tell that apart from a child the file simply omitted. Checking first keeps the answer all or nothing.

### Not covered here

- Only the schema → behavior mapping is capped. A file deep enough to overflow the YAML decoder itself fails earlier, in `OpenTelemetryConfigurationParser`, which is a separate problem.
- Compat mode still ignores declarative sampler config. That is the other open item on #914 and will come as its own PR, modelled on #1050.

I picked 10 as a limit no real config should ever reach. Happy to move it, or to report an over-deep file through `SdkErrorHandler` rather than silently unsetting the sampler — I raised both questions on the issue.

## Testing

`./gradlew :config-yaml:jvmTest detekt apiCheck`

Four cases added to `SamplerMapperTest`: nesting maps correctly at two levels, still maps at the limit, degrades to unset one level past it, and degrades to unset at 10,000 levels. Removing the new check and re-running makes that last one fail with `StackOverflowError`, which is the crash this fixes.


fractalwrench (approved):
LGTM - thanks!

PR #1072, opened 2026-09-18
Create propagators API module

fractalwrench:
## Goal

Create a new module named `api-propagators` that will be the entrypoint for obtaining instances of propagators without needing to rely on the `OpenTelemetry` interface. This will be utilised in the following way:

```kotlin
val propagators: Propagators = createPropagators()
propgators.none // obtain a propagator

createOpenTelemetry(
    propagators = propagators // optionally specify propgators to reuse as needed
)
```

This is a step towards supporting propagation separately from the OpenTelemetry interface.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

Issue #917, opened 2026-08-21
Configuration: SimpleSpanProcessor


fractalwrench:
## SimpleSpanProcessor

SimpleSpanProcessor: exports each span as it finishes, with no buffering.
Contains the following fields: `exporter` (required)

Declarative config path: `.tracer_provider.processors[].simple`
Schema type: `SimpleSpanProcessor`
Env vars: none

Spec links:

- Simple processor: https://opentelemetry.io/docs/specs/otel/trace/sdk/#simple-processor
- YAML schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/tracer_provider.yaml

Implementation checklist:

- [ ] Behavior implemented in `behavior` module?
- [ ] Readable from `config-dsl` module?
- [ ] Readable from `config-envar` module? (n/a)
- [ ] Readable from `config-yaml` module?
- [ ] Compat mode SDK obeys config
- [ ] Implementation mode SDK obeys config

aaravmahajanofficial:
Working on this.

Rajkaran-122:
@fractalwrench sir PTAL #1074 .

PR #1074, opened 2026-09-18
feat: add SimpleSpanProcessor configuration support

Rajkaran-122:
Implements behavior-based configuration for SimpleSpanProcessor:
- Updated TracerProviderConfigImpl to include processor in toBehavior()
- Added applyResolvedProcessor method to respect behavior-based processor config
- Updated config-dsl to map export() calls to console exporter behavior
- Updated compat mode to respect processor behavior from Java side
- Added tests for processor behavior resolution

This is a partial implementation that currently only supports console exporter via the behavior system. The full SimpleSpanProcessor issue requires additional work to support other exporter types.

Addresses #917


Rajkaran-122:
@fractalwrench sir PTAL .

Rxfa:
Following what was agreed [here](https://github.com/open-telemetry/opentelemetry-kotlin/pull/978#discussion_r3922949629), wouldn't it make more sense to start with a PR implementing the behavior for `simpleSpanProcessor` and leaving it readable from `config-envars`, `config-dsl` and `config-yaml` before jumping into doing the implementation?

Specially considering that we have nothing for `spanExporter` yet

Maybe we could aim to have something like the following for the behaviors of both `simpleSpanProcessor` and `batchSpanProcessor` as some sort of starting point?

```kotlin

@ExperimentalApi
sealed class SpanProcessorBehavior : Behavior<SpanProcessorBehavior> {
    data class BatchSpanProcessorBehavior(
        val console: ConsoleExporterBehavior? = null
    ){
        fun mergeWith(higher: BatchSpanProcessorBehavior): BatchSpanProcessorBehavior
    }

    data class SimpleSpanProcessorBehavior(
        val console: ConsoleExporterBehavior? = null
    ){
        fun mergeWith(higher: SimpleSpanProcessorBehavior): SimpleSpanProcessorBehavior
    }
}

```

fractalwrench:
I agree with the approach laid out by @Rxfa - it makes sense to implement the config for `SpanExporter` first before attempting `SpanProcessor`, and additionally it would be preferable to have it readable from the `config-envar/config-yaml/config-dsl` modules before wiring it up to control the SDK's behavior.

Rajkaran-122:
@fractalwrench sir Fixed the CI issue by updating `TracerProviderConfigImpl` from the old `processorBehavior?.console` API to the new `exporter`-based `SpanProcessorBehavior`.

Also updated the tests and handled both Simple and Batch processors. JVM tests pass.

I checked for remaining `SpanProcessorBehavior.console` references; none remain. The other `processorBehavior?.console` reference is for `LogRecordProcessorBehavior` in PR #1076, so I left it unchanged to keep this PR scoped.

Rajkaran-122:
@fractalwrench sir PTAL.

fractalwrench:
@Rajkaran-122 can you address the build failures? Thanks

behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/SpanExporterBehavior.kt
Rxfa: In the future, we may expect multiple exporters to be configured simultaneously
Isn't this kinda pushing towards a single-exporter model?

PR #1073, opened 2026-09-18
fix: move modified tracestate keys to the front

Rajkaran-122:
## Description

Fixes #1056

Updates `TraceStateImpl` to comply with the W3C Trace Context specification when modifying `tracestate` entries.

According to the specification:

* Modified keys **MUST** be moved to the beginning (left) of the list.
* New key/value pairs **SHOULD** be added to the beginning of the list.
* The relative order of unmodified key/value pairs **MUST** be preserved.

### Changes

* Changed `TraceStateImpl` to use `LinkedHashMap` to preserve `tracestate` entry ordering.
* Updated `put()` so modified or newly added keys are placed at the beginning.
* Updated `remove()` to correctly maintain ordering of the remaining entries.
* Updated `TraceStateMarshaller.decode()` to construct `TraceState` directly from the decoded map, preserving the original header order.
* Added/updated tests covering ordering behavior when entries are added, modified, and removed.

### Testing

Verified the expected `tracestate` ordering behavior through the updated test suite.

### Expected behavior

PR #1075, opened 2026-09-18
fix: TraceState.Put not moving modified keys to the front of the list

### LINKED ISSUE #1056: `TraceState.put` does not move modified keys to the front of the list
The [spec states](https://opentelemetry.io/docs/specs/otel/trace/api/#tracestate):

>When mutating `tracestate`, the order of unmodified key/value pairs MUST be preserved.
**Modified keys MUST be moved to the beginning (left) of the list.***

Currently `TraceStateImpl` appends the new key to the end of a `LinkedHashMap`, meaning the wire format receives the wrong order. `put` needs to remove any existing entry for the key and prepend the new paid, which leaves the correct order.

Rxfa:
Closes #1056 

## Goal

Make `TraceState.Put` respect the W3C spec, which states that unmodified key/value pairs must be preserved and modified keys must be added to the start of the list

## Testing

Added unit tests.


fractalwrench:
Closing in favour of #1073

Issue #963, opened 2026-08-24
JVM platform clock implementation improvements


fractalwrench:
opentelemetry-java anchors on `currentTimeMillis()` once and measures elapsed time
afterwards with `System.nanoTime()`. This would be a better approach for JVM (but not android).

For Android we should see the approach taken in opentelemetry-android, which baselines on `currentTimeMillis()`. Nanosecond resolution is not possible on Android, so millisecond resolution is acceptable.

I haven't looked into JS/Apple but it's possible improvements are available there too.

Shaurya2k06:
Looked at what we have vs Java / Android.

Today getCurrentTimeNanos() is currentTimeMillis() * 1_000_000 on both JVM and Android (same jvmAndAndroid source set), Date.now() on JS, and NSDate.timeIntervalSince1970 on Apple. Every Clock.now() is a fresh wall-clock read, we don’t have Java’s AnchoredClock (epoch + nanoTime() delta on the local root span).

Java is two layers: SystemClock.now() uses Instant (not millis), and span timestamps are AnchoredClock (clock.now() once, then clock.nanoTime()). Android’s OtelAndroidClock is the same idea but monotonic via SystemClock.elapsedRealtimeNanos() (covers deep sleep; System.nanoTime() doesn’t). GNSS/NTP anchors exist but are off by default; fallback is currentTimeMillis().

I’d split JVM off Android: JVM Instant + nanoTime() (or a small anchored clock for spans), Android currentTimeMillis() + elapsedRealtimeNanos(). JS can do Date.now() + performance.now(); Apple already has sub-ms from NSDate. Have a look, what do you think?

cc @fractalwrench

fractalwrench:
That seems like a reasonable initial approach. I think it'd be good to land all these in separate PRs that are independent of each other.

Rajkaran-122:
@fractalwrench sir PTAL #1077 .

fractalwrench:
Created #1083 to track the Android implementation separately, as I think this could benefit from being broken off into two issues.

PR #1077, opened 2026-09-19
perf: improve clock implementation with anchored clock

Rajkaran-122:
Implements clock performance improvements per issue #963:
- JVM: Use anchored clock pattern (anchor on currentTimeMillis, measure with nanoTime)
- Android: Keep millisecond resolution (acceptable per opentelemetry-android approach)
- Split jvmAndAndroidMain into separate jvm and android implementations
- Update tests to verify nanosecond precision on JVM

The anchored clock provides better timing precision by:
1. Anchoring epoch once using currentTimeMillis()
2. Measuring elapsed time with System.nanoTime() (higher resolution)
3. Periodically updating anchor to prevent overflow

This pattern follows opentelemetry-java's approach for improved timing.


Rajkaran-122:
@fractalwrench sir PTAL .

platform-implementations/src/androidMain/kotlin/io/opentelemetry/kotlin/ClockImpl.android.kt
```diff
@@ -1,5 +1,7 @@
 package io.opentelemetry.kotlin
 
 public actual fun getCurrentTimeNanos(): Long {
+    // Android can accept millisecond resolution per issue #963
+    // Following opentelemetry-android approach
```
fractalwrench, line 5:
```suggestion
```

platform-implementations/src/jvmMain/kotlin/io/opentelemetry/kotlin/ClockImpl.jvm.kt
```diff
@@ -0,0 +1,49 @@
...
+    fun getNanos(): Long {
+        val currentNanos = System.nanoTime()
+        val elapsedNanos = currentNanos - anchorNanos
+        
+        // Update anchor if we're getting close to overflow
+        // System.nanoTime() overflows every ~292 years, so we update every ~100 years
+        if (elapsedNanos > 3_155_760_000_000_000_000L) { // ~100 years in nanoseconds
+            updateAnchor()
+            return getNanos()
+        }
```
fractalwrench, line 43: IMO this is probably overkill

platform-implementations/src/jvmMain/kotlin/io/opentelemetry/kotlin/ClockImpl.jvm.kt
```diff
@@ -0,0 +1,49 @@
...
+    return getAnchoredClockNanos()
+}
+
+/**
+ * Anchored clock implementation that provides nanosecond precision by:
+ * 1. Anchoring the epoch once using currentTimeMillis()
+ * 2. Measuring elapsed time using System.nanoTime() (which has higher resolution)
+ * 3. Periodically updating the anchor to prevent overflow
+ * 
+ * This pattern is used by opentelemetry-java for better timing precision.
```
fractalwrench, line 16: opentelemetry-java also uses `SystemClock` which is the default implementation used: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/common/SystemClock.java

fractalwrench (commented):
I think it'd be good to scope this PR to only address the JVM clock implementation so I split the issue up accordingly. I'd like to understand more from @breedx-splk on how opentelemetry-java uses `SystemClock` and `AnchoredClock`, and whether there's any tradeoffs/context we might be missing, so let's wait to hear his thoughts

Issue #928, opened 2026-08-21
Configuration: OtlpHttpExporter


fractalwrench:
## OtlpHttpExporter

OtlpHttpExporter: OTLP over HTTP, for spans and log records.
Contains the following fields: `endpoint`, `headers` (list), `headers_list`, `compression`,
`timeout`, `encoding` (`protobuf` | `json`), `tls.{ca_file, cert_file, key_file}`

Declarative config path: `.{tracer_provider,logger_provider}.processors[].*.exporter.otlp_http`
Schema type: `OtlpHttpExporter`, `HttpTls`, `OtlpHttpEncoding`

Spec links:

- Configuration options: https://opentelemetry.io/docs/specs/otel/protocol/exporter/#configuration-options
- OTLP exporter: https://opentelemetry.io/docs/specs/otel/protocol/exporter/
- Env vars index: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#otlp-exporter
- YAML schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/common.yaml

Implementation checklist:

- [ ] Behavior implemented in `behavior` module?
- [ ] Readable from `config-dsl` module?
- [ ] Readable from `config-envar` module?
- [ ] Readable from `config-yaml` module?
- [ ] Compat mode SDK obeys config
- [ ] Implementation mode SDK obeys config

PR #1078, opened 2026-09-20
feat(behavior): Add OTLP HTTP exporter behavior for LogRecordProcesso…

Rxfa:
Part of https://github.com/open-telemetry/opentelemetry-kotlin/issues/928
This is the 1st of a small series of PRs

## Goal

- Added OTLP HTTP exporter to `spanProcessorBehavior` and `LogRecordProcessorBehavior`

Next PRs aim to make the HTTP exporter readable from `yaml`, `dsl` and `envar` configs.

## Testing

Added unit tests


fractalwrench (commented):
We only support a few options in the OTLP HTTP exporter right now. I think it'd probably be good to only add fields that we already support on `HttpExporterBehavior` (or open a separate PR that supports the new config options). Otherwise this approach looks good

https://github.com/open-telemetry/opentelemetry-kotlin/blob/b58f9fc56d5b6ff39c4f3661ad695d35ff0fabd1/exporters-otlp/src/commonMain/kotlin/io/opentelemetry/kotlin/export/OtlpHttpExporterConfigDsl.kt#L17

Rxfa:
> or open a separate PR that supports the new config options

Was going for this one, and then I saw that https://github.com/open-telemetry/opentelemetry-kotlin/issues/974 is open and has been assigned.

I'll make a change to only include the fields supported by `OtlpHttpExporterConfigDsl` and once @Shaurya2k06 adds support to the new configs, I'll expand the OTLP HTTP exporter

behavior/src/commonMain/kotlin/io/opentelemetry/kotlin/behavior/HttpExporterBehavior.kt
```diff
@@ -0,0 +1,90 @@
...
+
+import io.opentelemetry.kotlin.ExperimentalApi
+
+@ExperimentalApi
+/**
+ * Selecting the OTLP HTTP exporter.
+ *
+ * https://opentelemetry.io/docs/specs/otel/protocol/exporter/
+ */
+data class HttpExporterBehavior(
```
Rxfa: Will also change this to `otlpHttpExporterBehaviour` for the sake of consistency

fractalwrench (approved):
LGTM - thanks!

Rxfa:
Fixed the pipeline @fractalwrench

PR #1084, opened 2026-09-21
Move propagation/context interfaces to separate api-propagators module

fractalwrench:
## Goal

Moves several public APIs that relate to propagation/context management from the `api` module to `api-propagators`. `api` module depends on `api-propagators` and no packages have changed so from an existing user's POV this should be an invisible change.

Notably, `api-propagators` has no dependency on any other module so can be used independently from the `OpenTelemetry` interface. This is the first real step towards creating propagators/span context objects independently of `OpenTelemetry` objects, which is a requirement for instrumentation authors. Usage would look like the following:

```kotlin
// 1. obtain instance of the Propagators API, the entrypoint
val api: Propagators = createPropagators()

// 2. example for obtaining a no-op TextMapPropagator before OpenTelemetry is created
val noopPropagator: TextMapPropagator = api.none()

// 3. optionally create an OpenTelemetry instance using the pre-created propagators
val otel = createOpenTelemetry(propagators = api)

// 4. use the same propagation/context mechanisms that were previously created
val noopPropagator: TextMapPropagator = otel.none()
```

### Discussion

1. Is `Propagators` and `createPropagators()` a good name for the API entrypoint - any alternative suggestions?
2. Should `OpenTelemetry` extend `Propagators` in its hierarchy? This gets us backwards compatibility (`openTelemetry.spanContext.invalid`) but we may want different syntax
3. The `api-propagators` module will supply a fully working implementation. `compat` mode already has a working implementation from the opentelemetry-java SDK. Do we override that, or split up `api-propagators` to decorate the compat mode?
4. How should configurable behavior be handled? For example, if an instrumentation author assumes that thread-local storage should be used to store context, but an SDK user uses a custom storage solution, should both just use their separate mechanisms and accept potential consequences?
5. `otel.propagator` returns the `TextMapPropagator` configured by the SDK. We should probably rename it?

### Follow on steps

This PR only moves the interfaces. Subsequent changelogs will need to extract a lot of the actual implementations out of the `implementation` module (depending on the outcome of our discussions).

## Testing

Relied on existing test coverage and the example app compiling.


fractalwrench:
SIG discussion:

For point 3, the consensus was for `api-propagators` to have one implementation that is used for both the `compat` and `implementation` modules. This was seen as fairly low-risk compared to other parts of the SDK such as the tracing pipeline.

Context storage using different mechanisms was flagged as a possible blocker to these plans. We agreed for Jamie to draft a PR that demonstrates how to wrap opentelemetry-java in a way that makes it use the opentelemetry-kotlin context storage layer. Additionally, it was noted that exposing interfaces from `api-propagators` meant it would be possible to retroactively override the KMP implementation with Java decorators if needed.

jdvr (approved):

bidetofevil (approved):
LGTM. Per SIG discussion, as long as the Java compat still behaves as it does now wrt to how context works (i.e. thread local based), I'm ok with this separation. The alternative of making the Java compat for propagators work without a Java implementation feels more complicated than I'd like, and that use case doesn't seem to be genuinely needed.

carlosalberto:
> The api-propagators module will supply a fully working implementation. 

So I have mixed feelings about this. **Right now** it only defines a few interfaces with no implementation at all. What actual implementation you plan to have? I also have mixed feelings about why this cannot be in the API itself (specially given the API has a dependency on this package).

And more precisely, what would be the use case for this? Is there a use case where a user would need to use this _and_ not import the API? (**Disclaimer**: I'm a little bit out of the loop in case this is needed for the compat layer to keep working, so I will check that out so we can discuss more, etc).

IIRC, in the Java case, there's a similar package called `context`, on which the `api` package depends on, but that only defines the `Context`, `TextMapPropagator` and related pieces. I think that was done so users could use the `Context` logic _even_ without the OTel ecosystem, e.g. people not doing observability but needing to do propagation (both in-process and inter-process one). That was an explicit path they took at the time, unlike other SIGs, like Python, which (perfectly) bake the context in the API.

Issue #536, opened 2026-05-22
Implement environment variable parsing rules


fractalwrench:
## Env var parsing rules

Spec links:

- Implementation guidelines: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#implementation-guidelines
- Parsing empty value: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#parsing-empty-value
- Type-specific guidance: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#type-specific-guidance
- Boolean: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean
- Numeric: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#numeric
- Enum: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#enum
- Error handling policy: https://opentelemetry.io/docs/specs/otel/error-handling/

Implementation checklist:

- [ ] An empty or whitespace-only value MUST be treated as unset, not as an empty string
- [ ] Booleans are case-insensitive `true`; **everything else is false**, including `1`, `yes` and `on`
- [ ] Invalid numeric and enum values MUST be logged and otherwise ignored, i.e. fall back to the default

fractalwrench:
We need to check whether this is already implemented, and if not, implement it.

rastiehaiev:
@fractalwrench Could you please assign it to me? I would like to investigate and address it.

Msksgm:
Hi @rastiehaiev and @fractalwrench — sorry for not commenting here before starting work. I had started looking into this locally before the assignment, but I should have communicated earlier to avoid overlapping efforts.

I have a small change for the first checklist item: treating empty and whitespace-only environment variable values as unset, with tests, while preserving non-blank strings as-is.

If it would help, I'd be happy to share the patch or open a small PR for just that part. Otherwise, I'm happy to step back and leave it with you. Apologies again for the late coordination.

rastiehaiev:
Hey @Msksgm, thanks for offering. If you don’t mind, I’d like to take this issue through to completion, since I already have the full change implemented locally. I’m just reorganizing it into two smaller PRs and expect to open the first one tomorrow.

Msksgm:
@rastiehaiev 

Thanks for letting me know! Absolutely, please go ahead — I'll leave this issue with you. Thanks for taking care of it!

[...]

PR #1086, opened 2026-09-22
feat: Implement environment variable parsing rules - new environment variable reader [1/2]

rastiehaiev:
## Goal

Part of #536.

This PR introduces a new environment variable reader implementation under `config.envar.reader`.

The new `EnvVarReader` returns an `EnvVarReadResult`, distinguishing between valid, unset, and invalid values. Valid values may also carry a warning when a fallback was applied, such as an invalid Boolean value falling back to `false`.

`ReportingEnvVarReader` unwraps these results, applies the appropriate fallback, and forwards warnings through a callback.

The existing reader and its consumers remain unchanged in this PR. A follow-up PR will migrate the existing `*EnvVars` classes and `OpenTelemetryConfigReader` to the new implementation, then remove the old reader.

## Testing

Added common tests covering:

- Integer, Boolean, and string parsing.
- Empty and whitespace-only values.
- Invalid values and non-negative integer validation.
- Warning reporting and fallback behavior.


fractalwrench (approved):
LGTM - thanks!

PR #1088, opened 2026-09-22
fix: guard log processor enabled checks

MartinCajiao:
## Summary

Guard each `LogRecordProcessor.enabled` call through `SdkErrorHandler`. If a processor throws, report the user-code error and default that check to enabled so a log is not dropped because another processor failed. Preserve `CancellationException` propagation and the existing short-circuit behavior.

Adds tests for the fallback, short-circuiting, and cancellation behavior. Related to #791.

## Validation

- `:exporters-core:jvmTest --tests '*CompositeLogRecordProcessorTest'`


fractalwrench:
@MartinCajiao please sign the CLA then someone can take a look at this. Thanks

